PR #12, opened 2017-02-18
Modify sysDescription file location

taoyl-ms:
Use a snmp-specific location instead of historical shared location


stcheng:
could you reopen another one?

PR #19, opened 2017-03-28
Implement ipNetToMediaPhysAddress (ARP table)

qiluo-msft:


src/ax_interface/encodings.py
```diff
@@ -143,12 +143,12 @@ def size(self):
         return 4 + self.length + util.pad4(self.length)
 
     def __str__(self):
-        return self.string.decode('ascii')
+        return self.string.decode('utf-8')
```
pavel-shirshov: Why do we need this?

src/ax_interface/util.py
```diff
@@ -98,3 +98,10 @@ def mac_decimals(mac):
     """
     return tuple(int(h, 16) for h in mac.split(":"))
 
+def ip2tuple(ip):
```
pavel-shirshov: ipv4 only? probably it's better to use ip2tuple_v4

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -34,7 +34,14 @@ def get_index(if_name):
...
     """
-    match = re.match(SONIC_ETHERNET_RE_PATTERN, if_name.decode())
+    return get_index_from_str(if_name.decode())
+
+def get_index_from_str(if_name):
+    """
+    OIDs are 1-based, interfaces are 0-based, return the 1-based index
+    Ethernet N = N + 1
+    """
+    match = re.match(SONIC_ETHERNET_RE_PATTERN, if_name)
```
pavel-shirshov: Probably it's better to compile this regexp in __init__?
Also is it possible to extract this value from string using something like
if_name.replace('Ethernet', '') ?

src/sonic_ax_impl/mibs/ietf/rfc1213.py
```diff
@@ -41,6 +44,53 @@ class DbTables(int, Enum):
...
+        for entry in python_arptable.get_arp_table():
+            dev = entry['Device']
+            mac = entry['HW address']
+            ip = entry['IP address']
+
+            if_index = mibs.get_index_from_str(dev)
+            if if_index is None: continue
+
+            mactuple = mac_decimals(mac)
+            machex = ''.join([chr(b) for b in mactuple])
```
pavel-shirshov: machex = ''.join(chr(b) for b in mactuple) should work for you

src/ax_interface/encodings.py
qiluo-msft: Fix it and add test case.

---
In reply to: [108312790](https://github.com/Azure/sonic-snmpagent/pull/19#discussion_r108312790) [](ancestors = 108312790)

src/sonic_ax_impl/mibs/__init__.py
qiluo-msft: Let's keep it for readability and optimize if we find it on critical path. re library has the opportunity to hide the optimization without users changing code.

---
In reply to: [108313222](https://github.com/Azure/sonic-snmpagent/pull/19#discussion_r108313222) [](ancestors = 108313222)

src/sonic_ax_impl/mibs/ietf/rfc1213.py
```diff
@@ -41,6 +44,53 @@ class DbTables(int, Enum):
...
+            mac = entry['HW address']
+            ip = entry['IP address']
+
+            if_index = mibs.get_index_from_str(dev)
+            if if_index is None: continue
+
+            mactuple = mac_decimals(mac)
+            machex = ''.join(chr(b) for b in mactuple)
+            # if MAC is all zero
+            #if not any(mac): continue
```
pavel-shirshov: Should we remove this line? What to do if the mac is 0?

src/sonic_ax_impl/mibs/ietf/rfc1213.py
```diff
@@ -41,6 +44,53 @@ class DbTables(int, Enum):
...
+
+            iptuple = ip2tuple_v4(ip)
+
+            subid = (if_index,) + iptuple
+            self.arp_dest_map[subid] = machex
+            self.arp_dest_list.append(subid)
+
+            # print(subid, dev, mac, ip)
```
pavel-shirshov: Should we remove it?

src/sonic_ax_impl/mibs/ietf/rfc1213.py
qiluo-msft: The correct behavior of incomplete ARP entry (an IP without MAC) is not clear in the protocol. ref: [http://cric.grenoble.cnrs.fr/Administrateurs/Outils/MIBS/?oid=1.3.6.1.2.1.4.22.1.2](http://cric.grenoble.cnrs.fr/Administrateurs/Outils/MIBS/?oid=1.3.6.1.2.1.4.22.1.2)
Just keep the code here for further refinement, if needed.

---
In reply to: [108499363](https://github.com/Azure/sonic-snmpagent/pull/19#discussion_r108499363) [](ancestors = 108499363)

src/sonic_ax_impl/mibs/ietf/rfc1213.py
qiluo-msft: Removed

---
In reply to: [108499446](https://github.com/Azure/sonic-snmpagent/pull/19#discussion_r108499446) [](ancestors = 108499446)

PR #21, opened 2017-04-25
[sonic_ax_impl]: Add support of port channels.

oleksandrivantsiv:
- Add port channels support to RFC1213 (1.3.6.1.2.1.2) and
  RFC2863 (1.3.6.1.2.1.31.1) MIBs.
- Read oper state, admin state and MTU value from DB.

### DIFF
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -6,7 +6,10 @@ from swsssdk import SonicV2Connector
 from sonic_ax_impl import logger, _if_alias_map
 
 COUNTERS_PORT_NAME_MAP = b'COUNTERS_PORT_NAME_MAP'
+LAG_TABLE = b'LAG_TABLE'
+LAG_MEMBER_TABLE = b'LAG_MEMBER_TABLE'
 SONIC_ETHERNET_RE_PATTERN = "^Ethernet(\d+)$"
+SONIC_PORTCHANNEL_RE_PATTERN = "^PortChannel(\d+)$"
 APPL_DB = 'APPL_DB'
 ASIC_DB = 'ASIC_DB'
 COUNTERS_DB = 'COUNTERS_DB'
@@ -29,10 +32,27 @@ def lldp_entry_table(if_name):
     return b'LLDP_ENTRY_TABLE:' + if_name
 
 
+def if_entry_table(if_name):
+    """
+    :param if_name: given interface to cast.
+    :return: PORT_TABLE key.
+    """
+    return b'PORT_TABLE:' + if_name
+
+
+def lag_entry_table(lag_name):
+    """
+    :param lag_name: given lag to cast.
+    :return: LAG_TABLE key.
+    """
+    return b'LAG_TABLE:' + lag_name
+
+
 def get_index(if_name):
     """
     OIDs are 1-based, interfaces are 0-based, return the 1-based index
     Ethernet N = N + 1
+    PortChannel N = N + 1000
     """
     return get_index_from_str(if_name.decode())
 
@@ -40,11 +60,17 @@ def get_index_from_str(if_name):
     """
     OIDs are 1-based, interfaces are 0-based, return the 1-based index
     Ethernet N = N + 1
+    PortChannel N = N + 1000
     """
     match = re.match(SONIC_ETHERNET_RE_PATTERN, if_name)
     if match:
         n = match.group(1)
         return int(n) + 1
+
+    match = re.match(SONIC_PORTCHANNEL_RE_PATTERN, if_name)
+    if match:
+        n = match.group(1)
+        return int(n) + 1000
 
 
 def config(**kwargs):
@@ -107,3 +133,39 @@ def init_sync_d_interface_tables():
         if_alias_map = dict(zip(if_name_map.keys(), if_name_map.keys()))
 
     return db_conn, if_name_map, if_alias_map, if_id_map, oid_sai_map, oid_name_map
+
+
+def init_sync_d_lag_tables(db_conn):
+    """
+    Helper method. Connects to and initializes LAG interface maps for SyncD-connected MIB(s).
+    :param db_conn: database connector
+    :return: tuple(lag_name_if_name_map, if_name_lag_name_map, oid_lag_name_map)
+    """
+    db_conn.connect(APPL_DB)
+
+    lag_entries = db_conn.keys(APPL_DB, b"LAG_TABLE:*")
+
+    # { lag_name (SONiC) -> [ lag_members (if_name) ] }
+    # ex: { "PortChannel0" : [ "Ethernet0", "Ethernet4" ] }
+    lag_name_if_name_map = {}
+
+    # { if_name (SONiC) -> lag_name }
+    # ex: { "Ethernet0" : "PortChannel0" }
+    if_name_lag_name_map = {}
+
+    for lag_entry in lag_entries:
+        lag_name = lag_entry[len(b"LAG_TABLE:"):]
+        lag_members = db_conn.keys(APPL_DB, b"LAG_MEMBER_TABLE:%s:*" % lag_name)
+
+        def member_name_str(val, lag_name):
+            return val[len(b"LAG_MEMBER_TABLE:%s:" % lag_name):]
+
+        lag_name_if_name_map[lag_name] = [member_name_str(m, lag_name) for m in lag_members]
+        for key, val in lag_name_if_name_map.items():
+            if_name_lag_name_map[key] = val
+
+    # { OID -> lag_name (SONiC) }
+    oid_lag_name_map = {get_index(if_name): if_name for if_name in lag_name_if_name_map.keys()
+                        if get_index(if_name)}
+
+    return lag_name_if_name_map, if_name_lag_name_map, oid_lag_name_map
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -56,21 +56,22 @@ def get_index(if_name):
     """
     return get_index_from_str(if_name.decode())
 
+
 def get_index_from_str(if_name):
     """
     OIDs are 1-based, interfaces are 0-based, return the 1-based index
     Ethernet N = N + 1
     PortChannel N = N + 1000
     """
-    match = re.match(SONIC_ETHERNET_RE_PATTERN, if_name)
-    if match:
-        n = match.group(1)
-        return int(n) + 1
+    patterns = {
+        SONIC_ETHERNET_RE_PATTERN: 1,
+        SONIC_PORTCHANNEL_RE_PATTERN: 1000
+    }
 
-    match = re.match(SONIC_PORTCHANNEL_RE_PATTERN, if_name)
-    if match:
-        n = match.group(1)
-        return int(n) + 1000
+    for pattern, baseidx in patterns.items():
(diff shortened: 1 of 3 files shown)


src/sonic_ax_impl/mibs/ietf/rfc1213.py
```diff
@@ -126,6 +133,12 @@ def get_counter(self, sub_id, table_name):
         :param table_name: the redis table (either IntEnum or string literal) to query.
         :return: the counter for the respective sub_id/table.
         """
+        if sub_id in self.oid_lag_name_map:
+            counter_value = 0
+            for lag_member in self.lag_name_if_name_map[self.oid_lag_name_map[sub_id]]:
+                counter_value += self.get_counter(mibs.get_index(lag_member), table_name)
```
pavel-shirshov: What if the sum of the interface counters will be greater then 2^32?

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -29,23 +32,46 @@ def lldp_entry_table(if_name):
...
     OIDs are 1-based, interfaces are 0-based, return the 1-based index
     Ethernet N = N + 1
+    PortChannel N = N + 1000
     """
     match = re.match(SONIC_ETHERNET_RE_PATTERN, if_name)
     if match:
         n = match.group(1)
         return int(n) + 1
 
+    match = re.match(SONIC_PORTCHANNEL_RE_PATTERN, if_name)
```
pavel-shirshov: Probably it's better to make a loop over regexps with if_name.
regexps = {
  SONIC_ETHERNET_RE_PATTERN: 1
  SONIC_PORTCHANNEL_RE_PATTERN: 1000
}
If we want to add vlan or other interfaces later, we could just add more entry to table.

src/sonic_ax_impl/mibs/ietf/rfc1213.py
oleksandrivantsiv: It will be truncated to 32 bit in upper layer. Changed code to truncate it explicitly.

lguohan:
@qiluo-msft, canyou review?

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -29,22 +32,46 @@ def lldp_entry_table(if_name):
...
     Ethernet N = N + 1
+    PortChannel N = N + 1000
     """
-    match = re.match(SONIC_ETHERNET_RE_PATTERN, if_name)
-    if match:
-        n = match.group(1)
-        return int(n) + 1
+    patterns = {
+        SONIC_ETHERNET_RE_PATTERN: 1,
+        SONIC_PORTCHANNEL_RE_PATTERN: 1000
```
qiluo-msft: >1000 [](start = 38, length = 4)

Magic number. Give a name and some comment? May be '1' above is in the same condition.

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -107,3 +134,39 @@ def init_sync_d_interface_tables():
...
+    # ex: { "PortChannel0" : [ "Ethernet0", "Ethernet4" ] }
+    lag_name_if_name_map = {}
+
+    # { if_name (SONiC) -> lag_name }
+    # ex: { "Ethernet0" : "PortChannel0" }
+    if_name_lag_name_map = {}
+
+    for lag_entry in lag_entries:
+        lag_name = lag_entry[len(b"LAG_TABLE:"):]
+        lag_members = db_conn.keys(APPL_DB, b"LAG_MEMBER_TABLE:%s:*" % lag_name)
```
qiluo-msft: regex replace LAG_TABLE->LAG_MEMBER_TABLE
Is it more readable?

qiluo-msft: >b"LAG_MEMBER_TABLE:%s:*" % lag_name [](start = 44, length = 35)

regex replace LAG_TABLE->LAG_MEMBER_TABLE
Is it more readable?

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -107,3 +134,39 @@ def init_sync_d_interface_tables():
...
+    # { lag_name (SONiC) -> [ lag_members (if_name) ] }
+    # ex: { "PortChannel0" : [ "Ethernet0", "Ethernet4" ] }
+    lag_name_if_name_map = {}
+
+    # { if_name (SONiC) -> lag_name }
+    # ex: { "Ethernet0" : "PortChannel0" }
+    if_name_lag_name_map = {}
+
+    for lag_entry in lag_entries:
+        lag_name = lag_entry[len(b"LAG_TABLE:"):]
```
qiluo-msft: >lag_entry[len(b"LAG_TABLE:"):] [](start = 19, length = 30)

Split by ':' and get the second part?

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -107,3 +134,39 @@ def init_sync_d_interface_tables():
...
+        def member_name_str(val, lag_name):
+            return val[len(b"LAG_MEMBER_TABLE:%s:" % lag_name):]
+
+        lag_name_if_name_map[lag_name] = [member_name_str(m, lag_name) for m in lag_members]
+        for key, val in lag_name_if_name_map.items():
+            if_name_lag_name_map[key] = val
+
+    # { OID -> lag_name (SONiC) }
+    oid_lag_name_map = {get_index(if_name): if_name for if_name in lag_name_if_name_map.keys()
+                        if get_index(if_name)}
```
qiluo-msft: >get_index [](start = 27, length = 9)

get_index() twice?

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -107,3 +134,39 @@ def init_sync_d_interface_tables():
...
+    if_name_lag_name_map = {}
+
+    for lag_entry in lag_entries:
+        lag_name = lag_entry[len(b"LAG_TABLE:"):]
+        lag_members = db_conn.keys(APPL_DB, b"LAG_MEMBER_TABLE:%s:*" % lag_name)
+
+        def member_name_str(val, lag_name):
+            return val[len(b"LAG_MEMBER_TABLE:%s:" % lag_name):]
+
+        lag_name_if_name_map[lag_name] = [member_name_str(m, lag_name) for m in lag_members]
```
qiluo-msft: >m [](start = 75, length = 1)

Check m is valid. And it will make sure recursive get_counter() is safe.

qiluo-msft:
>             return counter_value

We can cache this return value, so if recursive happens, it can be fetch back immediately.

---
Refers to: src/sonic_ax_impl/mibs/ietf/rfc1213.py:152 in 61c87fb. [](commit_id = 61c87fbd3f11d2ce94367b00774eea821498147a, deletion_comment = False)

qiluo-msft:
>     def interface_name(self, sub_id):

I notice lots functions are the same as rfc1213. Suggest to refactor.

---
Refers to: src/sonic_ax_impl/mibs/ietf/rfc2863.py:73 in 61c87fb. [](commit_id = 61c87fbd3f11d2ce94367b00774eea821498147a, deletion_comment = False)

src/sonic_ax_impl/mibs/__init__.py
oleksandrivantsiv: There is comment in function description

src/sonic_ax_impl/mibs/__init__.py
oleksandrivantsiv: what is the difference?

src/sonic_ax_impl/mibs/__init__.py
oleksandrivantsiv: "m" is always valid we get it from DB

oleksandrivantsiv:
Refactor is not in scope of this commit. Suggest to create an enhancement

src/sonic_ax_impl/mibs/__init__.py
oleksandrivantsiv: It will look the same. Isn't it?

src/sonic_ax_impl/mibs/__init__.py
oleksandrivantsiv: Yes, to verify that get_index return value is not None

src/sonic_ax_impl/mibs/__init__.py
oleksandrivantsiv: Recursion here is safe. Maximum depth is always 1.

src/sonic_ax_impl/mibs/__init__.py
qiluo-msft: More readable.

---
In reply to: [113862724](https://github.com/Azure/sonic-snmpagent/pull/21#discussion_r113862724) [](ancestors = 113862724)

src/sonic_ax_impl/mibs/__init__.py
qiluo-msft: Could you optimize it to once?

---
In reply to: [113880393](https://github.com/Azure/sonic-snmpagent/pull/21#discussion_r113880393) [](ancestors = 113880393)

src/sonic_ax_impl/mibs/__init__.py
qiluo-msft: There is a trust boundary between DB and this component. If anything bad happens in DB, this component should not trapped in a non-stopping recursion and crash.

---
In reply to: [113863174](https://github.com/Azure/sonic-snmpagent/pull/21#discussion_r113863174) [](ancestors = 113863174)

oleksandrivantsiv: SNMP in SONiC works in read only mode. It takes information from DB converts it in special format and returns to user. Without working DB it is meaningless. If something bad happens with DB it will crash anyway because all code is implemented taking into consideration that DB can't fail. This is actual for any existing component.

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -107,3 +134,39 @@ def init_sync_d_interface_tables():
...
+
+def init_sync_d_lag_tables(db_conn):
+    """
+    Helper method. Connects to and initializes LAG interface maps for SyncD-connected MIB(s).
+    :param db_conn: database connector
+    :return: tuple(lag_name_if_name_map, if_name_lag_name_map, oid_lag_name_map)
+    """
+    db_conn.connect(APPL_DB)
+
+    lag_entries = db_conn.keys(APPL_DB, b"LAG_TABLE:*")
```
lguohan: this function seems to be only executed at the init time. this may not be complete at certain point of time. For example, the system has 4 port channels, and at the time when you tries to get the LAG_TABLE, you only see 3.

oleksandrivantsiv: Will update this PR soon

oleksandrivantsiv:
@qiluo-msft Created an issue to refactor rfc1213.py and rfc2863.py files https://github.com/Azure/sonic-snmpagent/issues/23. Will clarify with Liat if someone from MLNX team can contribute to this.

src/sonic_ax_impl/mibs/__init__.py
oleksandrivantsiv: done

qiluo-msft:
This merging breaks unit test. @oleksandrivantsiv Could you help check?
```
pytest -s -k tests/test_interfaces.py
```

PR #26, opened 2017-05-20
Implement ipRouteTable.ipRouteNextHop in MIB-2 (only default routes)

qiluo-msft:


src/sonic_ax_impl/mibs/ietf/rfc1213.py
```diff
@@ -84,8 +85,55 @@ def get_next(self, sub_id):
...
+
+        self.db_conn.connect(mibs.APPL_DB)
+        route_entries = self.db_conn.keys(mibs.APPL_DB, "ROUTE_TABLE:*")
+
+        for route_entry in route_entries:
+            routestr = route_entry.decode()
+            ipnstr = routestr[len("ROUTE_TABLE:"):]
+            if ipnstr == "0.0.0.0/0":
+                ipn = ipaddress.ip_network(ipnstr)
+                ent = self.db_conn.get_all(mibs.APPL_DB, routestr, blocking=True)
```
pavel-shirshov: if blocking is true you need to check that ent is not None. I'd make it unblocking

pavel-shirshov: I was wrong. It's better to have it blocking here :) Sorry

src/sonic_ax_impl/mibs/ietf/rfc1213.py
```diff
@@ -84,8 +85,55 @@ def get_next(self, sub_id):
...
+    def update_data(self):
+        """
+        Update redis (caches config)
+        Pulls the table references for each interface.
+        """
+        self.nexthop_map = {}
+        self.route_list = []
+
+        self.db_conn.connect(mibs.APPL_DB)
+        route_entries = self.db_conn.keys(mibs.APPL_DB, "ROUTE_TABLE:*")
```
pavel-shirshov: This could return None

PR #29, opened 2017-06-06
Exclude non front panel interfaces from nexthop

qiluo-msft:


lguohan:
I am not sure if this is the best approach, it is better to change fpmsyncd and not to put the route into app db in the first place.

stcheng:
then we need to draw a conclusion that APPL_DB will only store the information that will be synced to the ASIC. we will need future work to update `neighsyncd`, `intfsyncd`, etc. to remove the kernel related but not ASIC related stuffs.

lguohan:
I think getting a dedicate namespace for swss is the cleaning way to solve the problem. In this case, it will only see the front panel related routes, neighbors, interfaces.

PR #31, opened 2017-06-08
Exclude non front panel interfaces from nexthop

qiluo-msft:
Following https://github.com/Azure/sonic-snmpagent/pull/29, fix a syntax bug.

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc4292.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc4292.py
@@ -40,7 +40,7 @@ class RouteUpdater(MIBUpdater):
                     ## Ignore non front panel interfaces
                     ## TODO: non front panel interfaces should not be in APPL_DB at very beginning
                     ## This is to workaround the bug in current sonic-swss implementation
-                    if ifn == "eth0" || ifn == "lo" || ifn == "docker0": continue
+                    if ifn == "eth0" or ifn == "lo" or ifn == "docker0": continue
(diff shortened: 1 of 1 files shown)


src/sonic_ax_impl/mibs/ietf/rfc4292.py
```diff
@@ -35,7 +35,12 @@ def update_data(self):
                 ipn = ipaddress.ip_network(ipnstr)
                 ent = self.db_conn.get_all(mibs.APPL_DB, routestr, blocking=True)
                 nexthops = ent[b"nexthop"].decode()
-                for nh in nexthops.split(','):
+                ifnames = ent[b"ifname"].decode()
+                for nh, ifn in zip(nexthops.split(','), ifnames.split(',')):
```
stcheng: according to my observation, it is possible that the number of nexthops and the number of ifnames cannot match. it happens when the ifnames is `eth0` but no nexthops.
```
127.0.0.1:6379> hgetall ROUTE_TABLE:10.3.146.0/23
1) "nexthop"
2) ""
3) "ifname"
4) "eth0"
```

qiluo-msft: The code here tolerates this case. zip() just returns a list as long as the shorter one of nexthops and ifnames.

stcheng: if it shortens the input, then the above case will return empty list I think

stcheng: I see. no problem here.

PR #32, opened 2017-06-09
Add logging for updaters and MIBTable background task

qiluo-msft:
Sample syslog for a flawed updater:
```
Jun  9 06:53:25.0 str-s6000-on-5 ERR snmp-subagent [ax_interface] ERROR: MIBUpdater.start() caught an unexpected exception#012Traceback (most recent call last):#012
File "/usr/local/lib/python3.6/dist-packages/ax_interface/mib.py", line 45, in start#012    self.update_data()#012  File "/usr/local/lib/python3.6/dist-packages/sonic_ax_impl/mibs/ietf/rfc4292.py", line 38, in update_data#012    for route_entry in route_entries:#012TypeError: 'NoneType' object is not iterable
```

### DIFF
--- a/src/ax_interface/mib.py
+++ b/src/ax_interface/mib.py
@@ -43,10 +43,7 @@ class MIBUpdater:
             # run the background update task
             try:
                 self.update_data()
-            except (KeyboardInterrupt, SystemExit):
-                # They are exception thrown when Ctrl+C or sys.exit() happens
-                raise
-            except:
+            except Exception:
                 # Any other exception or error, log it and keep running
                 logger.exception("MIBUpdater.start() caught an unexpected exception")
 
@@ -240,7 +237,7 @@ class MIBTable(dict):
         self.updater_instances = getattr(mib_cls, MIBMeta.UPDATERS)
         self.prefixes = getattr(mib_cls, MIBMeta.PREFIXES)
 
-    def done_background_task_callback(fut):
+    def _done_background_task_callback(fut):
         ex = fut.exception()
         if ex is not None:
             exstr = "MIBTable background task caught an unexpected exception: {}".format(str(ex))
@@ -252,7 +249,7 @@ class MIBTable(dict):
             updater.frequency = self.update_frequency
             updater.run_event = event
             fut = asyncio.ensure_future(updater.start())
-            fut.add_done_callback(MIBTable.done_background_task_callback)
+            fut.add_done_callback(MIBTable._done_background_task_callback)
             task = event._loop.create_task(fut)
             tasks.append(task)
         return asyncio.gather(*tasks, loop=event._loop)
--- a/src/ax_interface/mib.py
+++ b/src/ax_interface/mib.py
@@ -1,8 +1,12 @@
+import sys
 import asyncio
 import bisect
 import random
+import logging
 
 from . import util
+from . import logger
+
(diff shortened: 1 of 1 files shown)


src/ax_interface/mib.py
```diff
@@ -37,7 +41,18 @@ def __init__(self):
...
+                self.update_data()
+            except (KeyboardInterrupt, SystemExit):
+                # They are exception thrown when Ctrl+C or sys.exit() happens
+                raise
+            except:
+                # Any other exception or error, log it and keep running
+                ex = sys.exc_info()[0]
+                exstr = "MIBUpdater.start() caught an unexpected exception: {}".format(str(ex))
+                print(exstr, flush=True)
+                logger.exception(exstr)
```
tombo315: `logger.exception` already captures the exc_info. What is here will print the exception, and have a logger message that will include the exception twice.
```python
logger.exception(exstr) 
```

Combining `print` with `logging` is a code smell.
If you want errors to go to stdout, ``addHandler(logging.StreamHandler(sys.stdout))` to https://github.com/Azure/sonic-snmpagent/blob/master/src/sonic_ax_impl/__main__.py#L39

src/ax_interface/mib.py
```diff
@@ -228,13 +243,22 @@ def __init__(self, mib_cls, update_frequency=DEFAULT_UPDATE_FREQUENCY):
         self.updater_instances = getattr(mib_cls, MIBMeta.UPDATERS)
         self.prefixes = getattr(mib_cls, MIBMeta.PREFIXES)
 
+    def done_background_task_callback(fut):
+        ex = fut.exception()
+        if ex is not None:
+            exstr = "MIBTable background task caught an unexpected exception: {}".format(str(ex))
+            print(exstr, flush=True)
+            logger.error(exstr)
```
tombo315: Same thing here.

src/ax_interface/mib.py
qiluo-msft: Already fixed in the 2nd commit. Could you double check?

---
In reply to: [121173665](https://github.com/Azure/sonic-snmpagent/pull/32#discussion_r121173665) [](ancestors = 121173665)

src/ax_interface/mib.py
```diff
@@ -228,13 +240,21 @@ def __init__(self, mib_cls, update_frequency=DEFAULT_UPDATE_FREQUENCY):
         self.updater_instances = getattr(mib_cls, MIBMeta.UPDATERS)
         self.prefixes = getattr(mib_cls, MIBMeta.PREFIXES)
 
+    def done_background_task_callback(fut):
+        ex = fut.exception()
+        if ex is not None:
+            exstr = "MIBTable background task caught an unexpected exception: {}".format(str(ex))
+            logger.error(exstr)
```
tombo315: I believe you want `exception`

qiluo-msft: I don't use 'logger.exception' here on purpose. All the updater's exceptions or errors should be caught by MIBUpdater, if anything caught here, it's really a logic error in the MIBUpdater or MIBTable itself, which is truly unexpected.

---
In reply to: [121222352](https://github.com/Azure/sonic-snmpagent/pull/32#discussion_r121222352) [](ancestors = 121222352)

PR #35, opened 2017-07-07
Reduce the log of invalid FDB_ENTRY

qiluo-msft:


stcheng:
@lguohan I think for orchagent, it is better to add oid mapping not only for port channels or ports, but all created swss objects. that could be helpful when debugging.

PR #36, opened 2017-07-10
Fix empty lag members, fix if_name_lag_name_map

qiluo-msft:
1. Fix init_sync_d_lag_tables(): supports empty LAG member
2. Fix init_sync_d_lag_tables(): if_name_lag_name_map behaves as comments
3. Add test case for LAG interfaces


src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -172,13 +172,16 @@ def init_sync_d_lag_tables(db_conn):
     for lag_entry in lag_entries:
         lag_name = lag_entry[len(b"LAG_TABLE:"):]
         lag_members = db_conn.keys(APPL_DB, b"LAG_MEMBER_TABLE:%s:*" % lag_name)
+        if lag_members is None:
```
lguohan: add TODO. db_conn.keys() should really return [] instead of None.

PR #37, opened 2017-09-21
[rfc4363]: Update FDB logics per change of the SAI

stcheng:
In SAI 1.0, the FDB entry no longer contains the port ID. Instead,
it contains the bridge port ID which then corelates to the port ID.


src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -34,12 +34,24 @@ def reinit_data(self):
...
 
+        self.if_bpid_map = {}
+        self.db_conn.connect(mibs.ASIC_DB)
+        bridge_port_strings = self.db_conn.keys(mibs.ASIC_DB, "ASIC_STATE:SAI_OBJECT_TYPE_BRIDGE_PORT:*")
+
+        if not bridge_port_strings:
+            return
+
+        for s in bridge_port_strings:
+            bridge_port_id = s[45:]
```
qiluo-msft: >45 [](start = 31, length = 2)

magic number

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -34,12 +34,24 @@ def reinit_data(self):
...
+            ent = self.db_conn.get_all(mibs.ASIC_DB, s, blocking=True)
+            port_id = ent[b"SAI_BRIDGE_PORT_ATTR_PORT_ID"][6:]
```
qiluo-msft: >6 [](start = 59, length = 1)

magic number

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -34,12 +34,24 @@ def reinit_data(self):
             self.prev_if_id_map = self.if_id_map
             self.invalid_port_oids = set()
 
+        self.if_bpid_map = {}
```
qiluo-msft: Add some comment for this block?

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -34,12 +34,24 @@ def reinit_data(self):
...
+            ent = self.db_conn.get_all(mibs.ASIC_DB, s, blocking=True)
+            port_id = ent[b"SAI_BRIDGE_PORT_ATTR_PORT_ID"][6:]
+            self.if_bpid_map[bridge_port_id] = port_id
+
     def update_data(self):
         """
         Update redis (caches config)
         Pulls the table references for each interface.
         """
-        self.db_conn.connect(mibs.ASIC_DB)
```
qiluo-msft: >connect [](start = 21, length = 7)

I think it is needed.

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -56,22 +68,11 @@ def update_data(self):
...
-            ## and should be ignored here
-            ## TODO: revisit logic here for LAG in VLAN
-            if port_oid not in self.if_id_map:
-                ## Reduce the logger amount by remember it in a set
-                if port_oid not in self.invalid_port_oids:
-                    self.invalid_port_oids.add(port_oid)
-                    mibs.logger.info("SyncD 'ASIC_DB' includes a FDB_ENTRY '{}' with an invalid interface id '{}'.".format(fdb_str, port_oid))
-                continue
+            bridge_port_id = ent[b"SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID"][6:]
+            port_id = self.if_bpid_map[bridge_port_id]
```
prsunny: Does it also cover the LAG case?

src/sonic_ax_impl/mibs/ietf/rfc4363.py
stcheng: resolved

src/sonic_ax_impl/mibs/ietf/rfc4363.py
stcheng: resolved

src/sonic_ax_impl/mibs/ietf/rfc4363.py
stcheng: resolved

src/sonic_ax_impl/mibs/ietf/rfc4363.py
stcheng: resolved

src/sonic_ax_impl/mibs/ietf/rfc4363.py
stcheng: good catch. I think right now we don't support LAG. I'll add todo here. We don't have such testbed setup in our env. I'll sync with Alibaba to check if they could help enhance this scenario.

PR #38, opened 2017-09-25
[sonic-ax-impl]: Add try/except logic to deal with KeyError for bridge port attribute

stcheng:


src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -48,8 +49,11 @@ def reinit_data(self):
             # Example output: ASIC_STATE:SAI_OBJECT_TYPE_BRIDGE_PORT:oid:0x3a000000000616
             bridge_port_id = s[45:]
             ent = self.db_conn.get_all(mibs.ASIC_DB, s, blocking=True)
-            port_id = ent[b"SAI_BRIDGE_PORT_ATTR_PORT_ID"][6:]
-            self.if_bpid_map[bridge_port_id] = port_id
+            try:
+                port_id = ent[b"SAI_BRIDGE_PORT_ATTR_PORT_ID"][6:]
+                self.if_bpid_map[bridge_port_id] = port_id
+            except KeyError:
+                logger.warning("SAI_OBJECT_TYPE_BRIDGE_PORT {} does not have SAI_BRIDGE_PORT_ATTR_PORT_ID attribute".format(bridge_port_id))
```
qiluo-msft: >logger.warning [](start = 16, length = 14)

mibs.logger, following other files in the same directory.

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -48,8 +49,11 @@ def reinit_data(self):
             # Example output: ASIC_STATE:SAI_OBJECT_TYPE_BRIDGE_PORT:oid:0x3a000000000616
             bridge_port_id = s[45:]
             ent = self.db_conn.get_all(mibs.ASIC_DB, s, blocking=True)
-            port_id = ent[b"SAI_BRIDGE_PORT_ATTR_PORT_ID"][6:]
-            self.if_bpid_map[bridge_port_id] = port_id
+            try:
+                port_id = ent[b"SAI_BRIDGE_PORT_ATTR_PORT_ID"][6:]
```
qiluo-msft: >SAI_BRIDGE_PORT_ATTR_PORT_ID [](start = 32, length = 28)

suggest check dict keys instead of try-catch.

src/sonic_ax_impl/mibs/ietf/rfc4363.py
qiluo-msft: >logger.warning [](start = 16, length = 14)

It will log the warning every min. Is it too much?

src/sonic_ax_impl/mibs/ietf/rfc4363.py
stcheng: `if key in dict` goes when it is a condition that we needs to deal with. since this is an abnormal situation, i think it fits better in `try/catch` scenario.

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -48,8 +49,11 @@ def reinit_data(self):
             # Example output: ASIC_STATE:SAI_OBJECT_TYPE_BRIDGE_PORT:oid:0x3a000000000616
             bridge_port_id = s[45:]
             ent = self.db_conn.get_all(mibs.ASIC_DB, s, blocking=True)
-            port_id = ent[b"SAI_BRIDGE_PORT_ATTR_PORT_ID"][6:]
-            self.if_bpid_map[bridge_port_id] = port_id
+            try:
```
qiluo-msft: Add some comments?

src/sonic_ax_impl/mibs/ietf/rfc4363.py
stcheng: this is abnormal, we need to debug on this situation.

src/sonic_ax_impl/mibs/ietf/rfc4363.py
stcheng: there shouldn't be scenarios that we have a bridge port without such attribute

stcheng:
thanks. i'll check if this is an issue or not in the SAI level before merging.

PR #39, opened 2017-09-26
Update the code to use common library API for bridge/port mapping

prsunny:


qiluo-msft:
Looks good to me. Did you run all the unit tests?

prsunny:
Yes, running. Will merge only after the execution.

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -62,6 +62,8 @@ def update_data(self):
             ent = self.db_conn.get_all(mibs.ASIC_DB, s, blocking=True)
             # Example output: oid:0x3a000000000608
             bridge_port_id = ent[b"SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID"][6:]
+            if bridge_port_id not in self.if_bpid_map:
```
qiluo-msft: I agree this is a workaround for current Broadcom SAI 1.0 implementation, however, the workaround is doing more than ignoring unexpected FDB entry. Instead, we should log this unexpected FDB entry, but not too verbose (every 5 sec).

After Broadcom fix the SAI behavior, we don't need this kind of workaround.

prsunny: Agree! will track the fix for unexpected FDB entry and revisit the work-around

PR #44, opened 2017-10-17
[rfc4364]: Add extra condition to deal with init scenario

stcheng:
When snmp subagent starts, the self.if_bpid_map is not initialized.
Add the extra condition to prevent NoneType from being iterated.

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc4363.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc4363.py
@@ -62,7 +62,7 @@ class FdbUpdater(MIBUpdater):
             ent = self.db_conn.get_all(mibs.ASIC_DB, s, blocking=True)
             # Example output: oid:0x3a000000000608
             bridge_port_id = ent[b"SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID"][6:]
-            if bridge_port_id not in self.if_bpid_map:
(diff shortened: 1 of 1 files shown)


qiluo-msft:
>         self.update_data()

actually it's better to reinit_data instead of update_data. Then you don't need to check basic data every time, as below changes.

---
Refers to: src/sonic_ax_impl/mibs/ietf/rfc4363.py:21 in 0159b88. [](commit_id = 0159b88d3c8b75c7443a78dfd3193343e0d8ca40, deletion_comment = False)

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -62,7 +62,7 @@ def update_data(self):
             ent = self.db_conn.get_all(mibs.ASIC_DB, s, blocking=True)
             # Example output: oid:0x3a000000000608
             bridge_port_id = ent[b"SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID"][6:]
-            if bridge_port_id not in self.if_bpid_map:
+            if not self.if_bpid_map or bridge_port_id not in self.if_bpid_map:
```
qiluo-msft: Should we fix get_bridge_port_map to return empty dict instead of None?

stcheng:
close this pull request in favor of https://github.com/Azure/sonic-py-swsssdk/pull/17

PR #49, opened 2017-10-30
[rfc4364]: Add extra condition to deal with init scenario

stcheng:
When snmp subagent starts, the self.if_bpid_map is not initialized.
Add the extra condition to prevent NoneType from being iterated.


src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -62,7 +62,7 @@ def update_data(self):
             ent = self.db_conn.get_all(mibs.ASIC_DB, s, blocking=True)
             # Example output: oid:0x3a000000000608
             bridge_port_id = ent[b"SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID"][6:]
-            if bridge_port_id not in self.if_bpid_map:
+            if not self.if_bpid_map or bridge_port_id not in self.if_bpid_map:
```
prsunny: Is if_bpid_map returned as {} in reinit_data() if the map is not populated during init?

stcheng: yes. thanks! it's fixed in the py-swsssdk repository!

stcheng:
already fixed in somewhere else.

PR #56, opened 2017-12-12
Implement ipCidrRouteStatus for default route and loopback

qiluo-msft:


tests/test_forward.py
```diff
@@ -137,3 +137,20 @@ def test_getnextpdu_empty(self):
...
+        )
+
+        encoded = get_pdu.encode()
+        response = get_pdu.make_response(self.lut)
+
+        n = len(response.values)
+        # self.assertEqual(n, 7)
+        value0 = response.values[0]
+        self.assertEqual(value0.type_, ValueType.INTEGER)
+        self.assertEqual(str(value0.name), '.1.3.6.1.2.1.4.24.4.1.16.0.0.0.0.0.0.0.0.0.10.0.0.1')
```
stcheng: `10.0.0.1` is not loopback IP?

PR #59, opened 2018-01-09
Fix the interface speed for snmp

zhenggen-xu:
Add interface speed support for snmp


lguohan:
can you add unit test for this one?

qiluo-msft:
In favor of https://github.com/Azure/sonic-snmpagent/pull/73

PR #60, opened 2018-01-17
Fix ipCidrRouteStatus for loopback: no need to check route table

qiluo-msft:
Signed-off-by: Qi Luo <qiluo-msft@users.noreply.github.com>


stcheng:
thanks

PR #61, opened 2018-01-23
Prevent integer overflow during PDU encoding

qiluo-msft:


qiluo-msft:
Thanks @tombo315 , understand your point.
My arguments are that there are different responsibility for caller and callee.
1. Caller has the counter value, and know how to present a out-of-range integer value, either saturating or wrapping, depending on the semantics of the counter. A good caller has already pre-limit the integer before calling callee.
2. Callee only cares about encoding, so it just protect itself from any exceptions. If caller provides out-of-range argument, the behavior is undefined. I simply choose 'wrapping' implementation in all possible behaviors. This is just prevent process crash, as we observed in production. Your proposal is also valid here.

The most callers implement wrapping logic. As more and more vendors contributing features coming, we need protection like this PR.

tombo315:
Protection, sure. 

I agree, ideally, this would never be needed. If you're seeing this in production, wouldn't this hide platform bugs that are reporting out-of-precision counters?

qiluo-msft:
We are still investigating the root cause and going to add unit test. The crash is super annoying than a bad represented counter in the production. We should rely on other measurement to expose the bug.

PR #62, opened 2018-01-29
Get Vlan Id from ASIC_STATE Vlan Object if bvid present

prsunny:
API reference : https://github.com/Azure/sonic-py-swsssdk/pull/30

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc4363.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc4363.py
@@ -9,10 +9,10 @@ from bisect import bisect_right
 
 def fdb_vlanmac(db_conn, fdb):
     if 'vlan' in fdb:
-        return (int(fdb["vlan"]),) + mac_decimals(fdb["mac"])
+        vlan_id = fdb["vlan"]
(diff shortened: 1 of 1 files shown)


src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -7,8 +7,12 @@
...
 from bisect import bisect_right
 
-def fdb_vlanmac(fdb):
-    return (int(fdb["vlan"]),) + mac_decimals(fdb["mac"])
+def fdb_vlanmac(db_conn, fdb):
+    if 'vlan' in fdb:
+        return (int(fdb["vlan"]),) + mac_decimals(fdb["mac"])
+    elif 'bvid' in fdb:
+        vlan_id = port_util.get_vlan_id_from_bvid(db_conn, fdb["bvid"])
+        return (int(vlan_id),) + mac_decimals(fdb["mac"])
```
yxieca: I don't feel this is a must-change;

I would assign vlan_id in both if and elif branches and return the vlanmac from a same statement, for the sake of consistency.

PR #69, opened 2018-07-03
Extend RFC 2737 MIB and implement RFC 3433

stepanblyschak:

### DIFF
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -29,6 +29,10 @@ SENSOR_PART_ID_MAP = {
     "tx2bias":      22,
     "tx3bias":      32,
     "tx4bias":      42,
+    "tx1power":     13,
+    "tx2power":     23,
+    "tx3power":     33,
+    "tx4power":     43,
 }
 
 # IfIndex to OID multiplier for transceiver
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -14,8 +14,8 @@ APPL_DB = 'APPL_DB'
 ASIC_DB = 'ASIC_DB'
 COUNTERS_DB = 'COUNTERS_DB'
 
-TABLE_NAME_SEPERATOR_COLON = ':'
-TABLE_NAME_SEPERATOR_VBAR = '|'
+TABLE_NAME_SEPARATOR_COLON = ':'
+TABLE_NAME_SEPARATOR_VBAR = '|'
 
 # This is used in both rfc2737 and rfc3433
 SENSOR_PART_ID_MAP = {
@@ -63,7 +63,7 @@ def transceiver_info_table(port_name):
     :return: transceiver info entry for this port
     """
 
-    return "TRANSCEIVER_INFO" + TABLE_NAME_SEPERATOR_VBAR + port_name
+    return "TRANSCEIVER_INFO" + TABLE_NAME_SEPARATOR_VBAR + port_name
 
 def transceiver_dom_table(port_name):
     """
@@ -71,7 +71,7 @@ def transceiver_dom_table(port_name):
     :return: transceiver dom entry for this port
     """
 
-    return "TRANSCEIVER_DOM_SENSOR" + TABLE_NAME_SEPERATOR_VBAR + port_name
+    return "TRANSCEIVER_DOM_SENSOR" + TABLE_NAME_SEPARATOR_VBAR + port_name
 
 def lldp_entry_table(if_name):
     """
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -14,6 +14,26 @@ APPL_DB = 'APPL_DB'
 ASIC_DB = 'ASIC_DB'
 COUNTERS_DB = 'COUNTERS_DB'
 
+TABLE_NAME_SEPERATOR_COLON = ':'
+TABLE_NAME_SEPERATOR_VBAR = '|'
+
+# This is used in both rfc2737 and rfc3433
+SENSOR_PART_ID_MAP = {
+    "temperature":  1,
+    "voltage":      2,
+    "rx1power":     11,
+    "rx2power":     21,
+    "rx3power":     31,
+    "rx4power":     41,
+    "tx1bias":      12,
+    "tx2bias":      22,
+    "tx3bias":      32,
+    "tx4bias":      42,
+}
+
+# IfIndex to OID multiplier for transceiver
+IFINDEX_SUB_ID_MULTIPLIER = 1000
+
 redis_kwargs = {'unix_socket_path': '/var/run/redis/redis.sock'}
 
 def counter_table(sai_id):
@@ -33,6 +53,21 @@ def queue_table(sai_id):
 def queue_key(port_index, queue_index):
     return str(port_index) + ':' + str(queue_index)
 
+def transceiver_info_table(port_name):
+    """
+    :param: port_name: port name
+    :return: transceiver info entry for this port
+    """
+
+    return "TRANSCEIVER_INFO" + TABLE_NAME_SEPERATOR_VBAR + port_name
+
+def transceiver_dom_table(port_name):
+    """
+    :param: port_name: port name
+    :return: transceiver dom entry for this port
+    """
+
+    return "TRANSCEIVER_DOM_SENSOR" + TABLE_NAME_SEPERATOR_VBAR + port_name
 
 def lldp_entry_table(if_name):
     """
@@ -224,3 +259,49 @@ def init_sync_d_queue_tables(db_conn):
 
     return port_queues_map, queue_stat_map, port_queue_list_map
 
+def get_device_metadata(db_conn):
+    """
+    :param db_conn: Sonic DB connector
+    :return: device metadata
+    """
+
+    DEVICE_METADATA = "DEVICE_METADATA|localhost"
+    db_conn.connect(db_conn.STATE_DB)
+
+    device_metadata = db_conn.get_all(db_conn.STATE_DB, DEVICE_METADATA)
+    return device_metadata
+
+def get_transceiver_sub_id(ifindex):
+    """
+    Returns sub OID for transceiver. Sub OID is calculated as folows:
+    +------------+------------+
+    |Interface   |Index       |
+    +------------+------------+
+    |Ethernet[X] |X * 1000    |
+    +------------+------------+
+    ()
+    :param ifindex: interface index
+    :return: sub OID of a port calculated as sub OID = {{index}} * 1000
+    """
+
+    return (ifindex * IFINDEX_SUB_ID_MULTIPLIER, )
+
+def get_transceiver_sensor_sub_id(ifindex, sensor):
+    """
+    Returns sub OID for transceiver sensor. Sub OID is calculated as folows:
+    +-------------------------------------+------------------------------+
+    |Sensor                               |Index                         |
+    +-------------------------------------+------------------------------+
+    |RX Power for Ethernet[X]/[LANEID]    |X * 1000 + LANEID * 10 + 1    |
+    |TX Bias for Ethernet[X]/[LANEID]     |X * 1000 + LANEID * 10 + 2    |
+    |Temperature for Ethernet[X]          |X * 1000 + 1                  |
+    |Voltage for Ethernet[X]/[LANEID]     |X * 1000 + 2                  |
+    +-------------------------------------+------------------------------+
+    ()
+    :param ifindex: interface index
+    :param sensor: sensor key
+    :return: sub OID = {{index}} * 1000 + {{lane}} * 10 + sensor id
+    """
+
+    transceiver_oid, = get_transceiver_sub_id(ifindex)
+    return (transceiver_oid + SENSOR_PART_ID_MAP[sensor], )
--- a/src/sonic_ax_impl/mibs/ietf/rfc2737.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc2737.py
@@ -1,33 +1,114 @@
+"""
+MIB implementation defined in RFC 2737
+"""
+
 from enum import Enum, unique
 from bisect import bisect_right
-import subprocess
 
-from swsssdk import SonicV2Connector
+from swsssdk import SonicV2Connector, port_util
 from ax_interface import MIBMeta, MIBUpdater, ValueType, SubtreeMIBEntry
 
+from sonic_ax_impl import mibs
 
 @unique
 class PhysicalClass(int, Enum):
     """
     Physical classes defined in RFC 2737.
     """
-    OTHER = 1
-    UNKNOWN = 2
-    CHASSIS = 3
-    BACKPLANE = 4
-    CONTAINER = 5
+
+    OTHER       = 1
+    UNKNOWN     = 2
+    CHASSIS     = 3
+    BACKPLANE   = 4
+    CONTAINER   = 5
     POWERSUPPLY = 6
-    FAN = 7
-    SENSOR = 8
-    MODULE = 9
-    PORT = 10
-    STACK = 11
+    FAN         = 7
+    SENSOR      = 8
+    MODULE      = 9
+    PORT        = 10
+    STACK       = 11
+
+
+@unique
+class XcvrInfoDB(bytes, Enum):
+    """
+    Transceiver info keys
+    """
+
+    TYPE              = b"type"
+    HARDWARE_REVISION = b"hardwarerev"
+    SERIAL_NUMBER     = b"serialnum"
+    MANUFACTURE_NAME  = b"manufacturename"
+    MODEL_NAME        = b"modelname"
+
+
+# Map used to generate sensor description 
+SENSOR_NAME_MAP = {
+    "temperature" : "Temperature",
+    "voltage"     : "Voltage",
+    "rx1power"    : "RX Power",
+    "rx2power"    : "RX Power",
+    "rx3power"    : "RX Power",
+    "rx4power"    : "RX Power",
+    "tx1bias"     : "TX Bias",
+    "tx2bias"     : "TX Bias",
+    "tx3bias"     : "TX Bias",
+    "tx4bias"     : "TX Bias",
+}
+
+
+def get_transceiver_data(xcvr_info):
+    """
+    :param xcvr_info: transceiver info dict
+    :return: tuple (type, hw_version, mfg_name, model_name) of transceiver;
+    Empty string if field not in xcvr_info
+    """
+
+    return (xcvr_info.get(xcvr_field.value, b"").decode()
+            for xcvr_field in XcvrInfoDB)
+
+
+def get_transceiver_description(sfp_type, if_alias):
+    """
+    :param sfp_type: SFP type of transceiver
+    :param if_alias: Port alias name
+    :return: Transceiver decsription
+    """
+
+    return "{} for {}".format(sfp_type, if_alias)
+
+def get_transceiver_sensor_description(sensor, if_alias):
+    """
+    :param sensor: sensor key name
+    :param if_alias: interface alias
+    :return: description string about sensor
+    """
+
+    # assume sensors that is per channel in transceiver port
+    # has digit equals to channel number in the sensor's key name in DB
+    # e.g. rx3power (lane 3)
+    lane_number = list(filter(lambda c: c.isdigit(), sensor))
+
+    if len(lane_number) == 0:
+        port_name = if_alias
+    elif len(lane_number) == 1 and int(lane_number[0]) in QSFP_LANES:
+        port_name = "{}/{}".format(if_alias, lane_number[0])
+    else:
+        mibs.logger.warning("Tried to parse lane number from sensor name - {} ".format(sensor)
+                + "but parsed value is not a valid QSFP lane number")
+        # continue as with non per channel sensor
+        port_name = if_alias
+
+    return "DOM {} Sensor for {}".format(SENSOR_NAME_MAP[sensor], port_name)
 
 
 class PhysicalTableMIBUpdater(MIBUpdater):
+    """
+    Updater class for physical table MIB
+    """
 
-    DEVICE_METADATA = "DEVICE_METADATA|localhost"
     CHASSIS_ID = 1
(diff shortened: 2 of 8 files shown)


qiluo-msft:
> CodeFlow for GitHub does not download executable files. If you trust this file and would like to run it, please download it yourself through GitHub.

Remove this file?

---
Refers to: src/ax_interface/.pdu_implementations.py.swp:1 in 0d42d47. [](commit_id = 0d42d47afd4baa86d045c6a43bce2f660297d5e4, deletion_comment = False)

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -1,32 +1,107 @@
...
+
+
+def get_transceiver_data(xcvr_info):
+    """
+    :param xcvr_info: transceiver info dict
+    :return: tuple (type, hw_version, mfg_name, model_name) of transceiver;
+    Empty string if field not in xcvr_info
+    """
+
+    return (xcvr_info.get(xcvr_field.value, b"").decode("utf-8")
```
qiluo-msft: >"utf-8" [](start = 56, length = 7)

You could just use default value.

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -1,32 +1,107 @@
...
+    """
+    :param sensor: sensor key name
+    :param if_alias: interface alias
+    :return: description string about sensor
+    """
+
+    # assume sensors that is per channel in transceiver port
+    # has digit equals to channel number in the sensor's key name in DB
+    digit = list(filter(lambda c: c.isdigit(), sensor))
+    if len(digit) == 1:
```
qiluo-msft: >digit [](start = 11, length = 5)

Are you assuming sensor number is 1-digit? Better remove relax it to multiple digits.

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -35,76 +110,256 @@ def __init__(self):
...
-        pass
+
+        chassis_sub_id = (self.CHASSIS_ID, )
+        self.physical_entities = [chassis_sub_id]
+
+        transceiver_key_pattern = mibs.transceiver_info_table("*")
+        self.transceiver_entries = self.statedb.keys(self.statedb.STATE_DB,
+                                                     transceiver_key_pattern)
+
+        if self.transceiver_entries is None:
```
qiluo-msft: >if self.transceiver_entries is None: [](start = 8, length = 36)

```if self.transceiver_entries:```
will work for both None and []

File: src/sonic_ax_impl/mibs/ietf/rfc2737.py [hunk 4]
```diff
@@ -35,76 +110,256 @@ def __init__(self):
...
+            # nothing in DB
+            return
+
+        # decode keys from DB
+        transceiver_entries = map(lambda k: k.decode("utf-8"), self.transceiver_entries)
```
qiluo-msft: >"utf-8" [](start = 53, length = 7)

the same.

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -35,76 +110,256 @@ def __init__(self):
...
+        # update xcvr info from DB
+        # use port's name as key for transceiver info entries
+        for transceiver_entry in transceiver_entries:
+            # extract interface name
+            interface = transceiver_entry.split("|")[-1]
+
+            ifalias = self.if_alias_map.get(interface.encode("utf-8"), b"")
```
qiluo-msft: >"utf-8" [](start = 61, length = 7)

the same

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -35,76 +110,256 @@ def __init__(self):
...
+            # decode alias
+            ifalias = ifalias.decode("utf-8")
```
qiluo-msft: >"utf-8" [](start = 37, length = 7)

the same

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -35,76 +110,256 @@ def __init__(self):
...
+
+            # get interface from interface name
+            ifindex = port_util.get_index_from_str(interface)
+
+            if ifindex is None:
+                # interface name invalid, skip this entry
```
qiluo-msft: Shall we log an error?

tests/mock_tables/state_db.json
```diff
@@ -1,5 +1,24 @@
 {
   "DEVICE_METADATA|localhost": {
     "chassis_serial_number": "SAMPLETESTSN"
+  },
+  "TRANSCEIVER_INFO|Ethernet0": {
+  	"type": "QSFP+",
+  	"hardwarerev" : "A1",
+  	"serialnum": "WW5062F",
+  	"manufacturename": "FINISAR CORP",
```
qiluo-msft: >FINISAR CORP [](start = 23, length = 12)

Prefer vendor neutral strings in test.

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -35,76 +110,256 @@ def __init__(self):
...
-                }
 
+        self.physical_classes_map[chassis_sub_id] = PhysicalClass.CHASSIS
+        self.physical_serial_number_map[chassis_sub_id] = chassis_serial_number
+
+        # update interface maps
+        _, self.if_alias_map, _, _, _ = \
+            mibs.init_sync_d_interface_tables(SonicV2Connector())
+            
     def update_data(self):
```
qiluo-msft: >update_data [](start = 8, length = 11)

This function is time critical. Try:
1. move all constants to __init__: like transceiver_key_pattern
2. move all semi constants to reinit_data: like transceiver_entries

src/sonic_ax_impl/mibs/ietf/rfc3433.py
```diff
@@ -0,0 +1,402 @@
...
+            # since sensor is unavailable
+            # vlaue can be 0
+            value = 0
+            oper_status = EntitySensorStatus.UNAVAILABLE
+        else:
+            # else the status is considered to be OK
+            oper_status = EntitySensorStatus.OK
+
+            value = value * 10 ** precision
+            if value >= EntitySensorValueRange.MAX:
```
qiluo-msft: >>= [](start = 21, length = 2)

>

src/sonic_ax_impl/mibs/ietf/rfc3433.py
```diff
@@ -0,0 +1,402 @@
...
+                value = EntitySensorValueRange.MAX
+            elif value <= EntitySensorValueRange.MIN:
```
qiluo-msft: ><= [](start = 23, length = 2)

<

src/sonic_ax_impl/mibs/ietf/rfc3433.py
```diff
@@ -0,0 +1,402 @@
...
+        is needed to convert value from one unit to anther
+        :return: value converted for MIB
+        """
+
+        type_ = cls.TYPE
+        scale = cls.SCALE
+        precision = cls.PRECISION
+
+        try:
+            value = converter(float(raw_value))
```
qiluo-msft: >converter [](start = 20, length = 9)

if converter raises a ValueError, should we treat it as a bug?

src/sonic_ax_impl/mibs/ietf/rfc3433.py
```diff
@@ -0,0 +1,402 @@
...
+    @classmethod
+    def mib_values(cls, raw_value):
+        """
+        NOTE: Here we convert rx power value from dB to mW.
+        Actually, rx power value in transceiver EEPROM is in watts,
+        but the value in STATE DB is in dB, so converting back.
+        """
+
+        # dBm -> mW convertion
+        converter = lambda x: 10 ** (x/10)
```
qiluo-msft: >converter [](start = 8, length = 9)

conv_dBm_mW?

src/sonic_ax_impl/mibs/ietf/rfc3433.py
```diff
@@ -0,0 +1,402 @@
...
+
+        # clear cache
+        self.ent_phy_sensor_type_map = {}
+        self.ent_phy_sensor_scale_map = {}
+        self.ent_phy_sensor_precision_map = {}
+        self.ent_phy_sensor_value_map = {}
+        self.ent_phy_sensor_oper_state_map = {}
+
+        transceiver_dom_keys = mibs.transceiver_dom_table("*")
+        self.transceiver_dom = self.statedb.keys(self.statedb.STATE_DB,
```
qiluo-msft: >transceiver_dom [](start = 13, length = 15)

just decode them?

src/sonic_ax_impl/mibs/ietf/rfc3433.py
```diff
@@ -0,0 +1,402 @@
...
+        for transceiver_dom_entry in transceiver_dom:
+            # extract interface name
+            interface = transceiver_dom_entry.split("|")[-1]
+            ifindex = port_util.get_index_from_str(interface)
+
+            if ifindex is None:
+                continue
+
+            # get transceiver sensors from transceiver dom entry in STATE DB
+            transceiver_dom = self.statedb.get_all(self.statedb.STATE_DB,
```
qiluo-msft: >transceiver_dom [](start = 12, length = 15)

why you change loop range inside a loop?

src/sonic_ax_impl/mibs/ietf/rfc2737.py
stepanblyschak: No, not a sensor number. Usually sensor name has a lane number in its name. So I think assuming lane number is 1 digit is ok, otherwise - it is something different

src/sonic_ax_impl/mibs/ietf/rfc2737.py
stepanblyschak: 1. Done, transceiver_key_pattern now is a constant class member
2. I agree that data in transceiver_entries like serial_num, vendor name should not change until we plug out the cable(which is quite rare), so having it in update_data() routine is not a best idea, however if I place it in reinit_data() we will receive changes about cable plugged/unplugged only every 1 min. On the other side, sensor values are dynamic data, so they are updated in update_data() every 5 sec. So we will have an interval when tables are not in sync. (E.g Ethernet0 is plugged out - 1 min there is still Ethernet0 in entities MIB, however Ethernet0 temperature sensor is not present in sensor MIB anymore)
Another way I thought about is that we can fetch the list of transceivers from DB in update_data(). If fetched list is the same as in last update_data() call - no update is needed, otherwise update cached values. In this case there is also an issue - If plug in and plug out happens in this 5 sec interval between update_data() calls we will not capture any changes and will have wrong values in cache until next reinit_data() call.

src/sonic_ax_impl/mibs/ietf/rfc3433.py
stepanblyschak: If float() raises ValueError than we cannot parse the value(It could be "N/A" in DB), then we teat sensor as UNAVAILABLE.
If converter raises a ValueError - we cannot convert it, (e.g if sensor value in DB is 0 and MIB should return this value in logarithmic scale, then converter will raise a ValueError and we will treat the sensor as UNAVAILABLE)
E.g:
>>> math.log(0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: math domain error

src/sonic_ax_impl/mibs/ietf/rfc3433.py
stepanblyschak: Sorry for that bad var naming. Agree this is confusing. Actually didn't noticed it because python allows have the same variable name in the inner scope without overriding outer scope variable. Now i fixed it

src/sonic_ax_impl/mibs/ietf/rfc3433.py
stepanblyschak: Yes, now changed it to decode right in place

src/sonic_ax_impl/mibs/ietf/rfc3433.py
stepanblyschak: Done.

tests/mock_tables/state_db.json
stepanblyschak: Done

src/sonic_ax_impl/mibs/ietf/rfc2737.py
stepanblyschak: Added a warning message, since we can skip the incorrect entry and continue.

src/sonic_ax_impl/mibs/ietf/rfc2737.py
stepanblyschak: Done

src/sonic_ax_impl/mibs/ietf/rfc2737.py
stepanblyschak: Done

src/sonic_ax_impl/mibs/ietf/rfc2737.py
stepanblyschak: Done

src/sonic_ax_impl/mibs/ietf/rfc2737.py
stepanblyschak: Done

src/sonic_ax_impl/mibs/ietf/rfc2737.py
stepanblyschak: Done

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -1,32 +1,107 @@
+"""
+MIB implementation defined in RFC 2737
+"""
+
 from enum import Enum, unique
 from bisect import bisect_right
-import subprocess
 
 from swsssdk import SonicV2Connector
+from swsssdk import port_util
```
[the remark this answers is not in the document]
stepanblyschak: Done

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -1,110 +1,366 @@
...
+    :return: description string about sensor
+    """
+
+    # assume sensors that is per channel in transceiver port
+    # has digit equals to channel number in the sensor's key name in DB
+    # e.g. rx3power (lane 3)
+    digit = list(filter(lambda c: c.isdigit(), sensor))
+    if len(digit) == 1:
+        port_name = "{}/{}".format(if_alias, *digit)
+    else:
```
qiluo-msft: >else [](start = 4, length = 4)

so if len() > 1, it should be an error.

src/sonic_ax_impl/mibs/ietf/rfc2737.py
qiluo-msft: Thanks for the thoughtful analysis. I prefer your 2nd, please keep explorer a better way to poll events, it will lead to considerable CPU optimization.

This optimization would benefit many other OIDs as well.

---
In reply to: [200943979](https://github.com/Azure/sonic-snmpagent/pull/69#discussion_r200943979) [](ancestors = 200943979)

src/sonic_ax_impl/mibs/ietf/rfc3433.py
qiluo-msft: I think the 'logarithmic' convert is a perfect example for my argument. If the raw value is 0, we should treat as sensor status available, but converter is a buggy one. We don't want to treat it as same as sensor status missing. Log it as a bug or propagate the error to downstream will help expose the bug.

---
In reply to: [200947198](https://github.com/Azure/sonic-snmpagent/pull/69#discussion_r200947198) [](ancestors = 200947198)

src/sonic_ax_impl/mibs/ietf/rfc2737.py
stepanblyschak: Added a warning if parsed lane_number is not valid QSFP lane

src/sonic_ax_impl/mibs/ietf/rfc2737.py
stepanblyschak: I refactored the code in update_data(). reinit_data() fully reinits the cache. In update_data() it listens to changes in STATE DB TRANSCEIVER_INFO table and updates only in case of change. In case of SET event we update only one entry, in case of DELETE event we remove this transceiver from available OIDs (not clearing the cache, cache will be refreshed in next reinit_data() call anyway).
pubsub.get_message() is non-blocking and also returns all unconsumed changes, so I'm not skipping events.

src/sonic_ax_impl/mibs/ietf/rfc3433.py
stepanblyschak: In case converter() call raises exception (ValueError or ArithmeticError) it is propaget to update_data() and an error is logged, entry skipped (since we don't know what to put into MIB, 0?)
I believe a good converter implementation for logarithmic convert should return float('-inf') in case original value is 0. In this case the code will correctly compare float('-inf') < EntitySensorValueRange.MIN, and set the value for mib to EntitySensorValueRange.MIN, so user will know that this sensor overflows and that this sensor is in logarithmic scale (according to TYPE field in MIB). Does it make sense?

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -1,110 +1,431 @@
...
+    :param if_alias: interface alias
+    :return: description string about sensor
+    """
+
+    # assume sensors that is per channel in transceiver port
+    # has digit equals to channel number in the sensor's key name in DB
+    # e.g. rx3power (lane 3)
+    lane_number = list(filter(lambda c: c.isdigit(), sensor))
+
+    if lane_number:
```
qiluo-msft: >if lane_number: [](start = 4, length = 15)

```if len(lane_number) == 0:```
will be much clear.

File: src/sonic_ax_impl/mibs/ietf/rfc2737.py [hunk 19]
```diff
@@ -1,110 +1,431 @@
...
+        # does not support pubsub
+        if not self.pubsub:
+            redis_client = self.statedb.get_redis_client(self.statedb.STATE_DB)
+            db = self.statedb.db_map[self.statedb.STATE_DB]["db"]
+            self.pubsub = redis_client.pubsub()
+            self.pubsub.psubscribe("__keyspace@{}__:{}".format(db, self.TRANSCEIVER_KEY_PATTERN))
+
+        while True:
+            msg = self.pubsub.get_message()
+            if msg:
```
qiluo-msft: >if msg: [](start = 12, length = 7)

suggest simple refactoring:
exit early if there is no msg.
so the remaining code could be one level less indented.

src/sonic_ax_impl/mibs/ietf/rfc3433.py
qiluo-msft: Make sense. Thanks for fixing and explanation.

---
In reply to: [201421302](https://github.com/Azure/sonic-snmpagent/pull/69#discussion_r201421302) [](ancestors = 201421302)

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -14,6 +14,30 @@
 ASIC_DB = 'ASIC_DB'
 COUNTERS_DB = 'COUNTERS_DB'
 
+TABLE_NAME_SEPERATOR_COLON = ':'
```
pavel-shirshov: typo SEPERATOR -> SEPARATOR: here and below

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -1,110 +1,435 @@
...
+        self.physical_classes_map[chassis_sub_id] = PhysicalClass.CHASSIS
+        self.physical_serial_number_map[chassis_sub_id] = chassis_serial_number
+
+        # retrieve the initial list of transceivers that are present in the system
+        transceiver_info = self.statedb.keys(self.statedb.STATE_DB, self.TRANSCEIVER_KEY_PATTERN)
+        if transceiver_info:
+            self.transceiver_entries = [entry.decode() \
+                for entry in transceiver_info]
+        else:
+            self.transceiver_entries = []
```
pavel-shirshov: I'd suggest to write it as
```
self.transceiver_entries = [entry.decode() for entry in transceiver_info]
```

you don't need if else here

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -1,110 +1,435 @@
...
+
+    def get_phy_serial_num(self, sub_id):
+        """
+        :param sub_id: sub OID
+        :return: serial number for this OID
+        """
+
+        if sub_id in self.physical_entities:
+            return self.physical_serial_number_map.get(sub_id, "")
+        return None
```
pavel-shirshov: The style is not so consistent:
```
return "" if sub_id in self.physical_entities else None 
```
vs
```
      if sub_id in self.physical_entities:  
            return self.physical_serial_number_map.get(sub_id, "") 
      return None 
```

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -1,110 +1,435 @@
...
+
+        return "" if sub_id in self.physical_entities else None
+
+    def get_phy_sw_rev(self, sub_id):
+        """
+        :param sub_id: sub OID
+        :return: software version for this OID
+        """
+
+        return "" if sub_id in self.physical_entities else None
```
pavel-shirshov: it returns empty string or None. It is right?

src/sonic_ax_impl/mibs/ietf/rfc2737.py
stepanblyschak: transceiver_info could be None so I can get exception "NoneType object is not iterable".
So maybe write `if transceiver_info is not None` so it will be clear why we need that if block?

src/sonic_ax_impl/mibs/__init__.py
pavel-shirshov: can you please fix the typo and I'll approve it?

src/sonic_ax_impl/mibs/ietf/rfc2737.py
stepanblyschak: Done, thanks

PR #70, opened 2018-07-04
[lldp] add new OIDs - lldpRemTable & lldpLocPortTable

mykolaf:
What I did:
Add support for lldpLocPortTable and lldpRemTable OIDs:

Also requires merge of https://github.com/Azure/sonic-dbsyncd/pull/5

OID | SNMP counter | Description
-- | -- | --
1.0.8802.1.1.2.1.3.7 | lldpLocPortTable |  
1.0.8802.1.1.2.1.3.7.1 | lldpLocPortEntry |  
1.0.8802.1.1.2.1.3.7.1.1 | lldpLocPortNum | The index value used to identify the port component (contained in the local chassis with the LLDP agent) associated with this entry
1.0.8802.1.1.2.1.3.7.1.2 | lldpLocPortIdSubtype | The type of port identifier encoding used in the associated 'lldpLocPortId' object
1.0.8802.1.1.2.1.3.7.1.4 | lldpLocPortDesc | The string value used to identify the 802 LAN station's port description associated with the local system. If the local agent supports IETF RFC 2863, lldpLocPortDesc object should have the same value of ifDescr object

OID | SNMP counter | Description
-- | -- | --
1.0.8802.1.1.2.1.4.1 | lldpRemTable |  
1.0.8802.1.1.2.1.4.1.1 | lldpRemEntry |  
1.0.8802.1.1.2.1.4.1.1.2 | lldpRemLocalPortNum | The index value used to identify the port component associated with this entry. The lldpRemLocalPortNum identifies the port on which the remote system information is received
1.0.8802.1.1.2.1.4.1.1.3 | lldpRemIndex | This object represents an arbitrary local integer value used by this agent to identify a particular connection instance, unique only for the indicated remote system
1.0.8802.1.1.2.1.4.1.1.11 | lldpRemSysCapSupported | The bitmap value used to identify which system capabilities are supported on the remote system
1.0.8802.1.1.2.1.4.1.1.12 | lldpRemSysCapEnabled | The bitmap value used to identify which system capabilities are enabled


tests/test_lldp.py
```diff
@@ -14,6 +14,7 @@
 from ax_interface.mib import MIBTable
 from sonic_ax_impl.mibs import ieee802_1ab
 
+import tests.mock_tables.dbconnector
```
qiluo-msft: >import [](start = 0, length = 6)

Move 'import' to after line ```from unittest import TestCase```, so it will not be override by later direct/indirect import.

tests/test_lldp.py
```diff
@@ -22,9 +23,14 @@ class LLDPMIB(ieee802_1ab.LLDPRemTable, ieee802_1ab.LLDPLocPortTable):
...
 
         cls.lut = MIBTable(LLDPMIB)
+        for updater in cls.lut.updater_instances:
+            updater.update_data()
+            updater.reinit_data()
+            updater.update_data()
 
     def test_getnextpdu_eth1(self):
         # oid.include = 1
+        # import pdb; pdb.set_trace()
```
qiluo-msft: >#  [](start = 8, length = 2)

remove commented code.

tests/mock_tables/appl_db.json
```diff
@@ -1,333 +1,675 @@
 {
-  "LLDP_ENTRY_TABLE:Ethernet100": {
```
qiluo-msft: >   [](start = 0, length = 2)

keep original indentation size?

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -5,25 +5,160 @@
...
+        """
+        Subclass update data routine. Updates available local port data.
+        """
+        # establish connection to application database.
+        self.db_conn.connect(mibs.APPL_DB)
+        self.if_range = []
+        # get local port kvs from APP_BD's PORT_TABLE
+        self.loc_port_data = {}
+        for if_oid, if_name in self.oid_name_map.items():
+            loc_port_kvs = self.db_conn.get_all(mibs.APPL_DB, mibs.if_entry_table(if_name))
```
qiluo-msft: >get_all [](start = 40, length = 7)

Do we need to update so frequently?

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -5,25 +5,160 @@
...
+        self.db_conn = mibs.init_db()
+        self.loc_chassis_data = {}
+
+    def update_data(self):
+        """
+        Subclass update data routine.
+        """
+        # establish connection to application database.
+        self.db_conn.connect(mibs.APPL_DB)
+        self.loc_chassis_data = self.db_conn.get_all(mibs.APPL_DB, mibs.LOC_CHASSIS_TABLE)
```
qiluo-msft: >get_all [](start = 45, length = 7)

Again, do we need to update so frequently?

mykolaf: You are right,  here the data is basically static, I will move this to reinit(), so it will be updated once a minute.

src/sonic_ax_impl/mibs/ieee802_1ab.py
mykolaf: I will subscribe for updates here.

PR #71, opened 2018-07-19
Fix/enhance lldp MIBs., add MockPubSub to mock_tables.dbconnector.py

mykolaf:
Add mock pubsub, so unittests can properly test MIBs with updaters that use keyspace notifications.
Also fix/enhance lldp MIBs.


src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -119,6 +121,9 @@ def update_data(self):
 
             # extract interface name
             interface = lldp_entry.split('|')[-1]
+            #ignore management interface
```
qiluo-msft: Add one blank

mykolaf: Checked with flake8

PR #72, opened 2018-07-20
[lldp] add lldpRemManAddrTable & lldpLocManAddrTable

mykolaf:
What I did:
- Implemented lldpRemManAddrTable & lldpLocManAddrTable

Also requires https://github.com/Azure/sonic-dbsyncd/pull/8

**lldpRemManAddrTable**

OID | SNMP counter | Description
-- | -- | --
1.0.8802.1.1.2.1.4.2.1.1 | lldpRemManAddrSubtype | The type of management address identifier encoding used in the associated 'lldpRemManagmentAddr' object
1.0.8802.1.1.2.1.4.2.1.2 | lldpRemManAddr | The string value used to identify the management address component associated with the remote system. The purpose of this address is to contact the management entity
1.0.8802.1.1.2.1.4.2.1.3 | lldpRemManAddrIfSubtype | The enumeration value that identifies the interface numbering method used for defining the interface number, associated with the remote system
1.0.8802.1.1.2.1.4.2.1.4 | lldpRemManAddrIfId | The integer value used to identify the interface number regarding the management address component associated with the remote system
1.0.8802.1.1.2.1.4.2.1.5 | lldpRemManAddrOID | The OID value used to identify the type of hardware component or protocol entity associated with the management address advertised by the remote system agent

**lldpLocManAddrTable**

OID | SNMP counter | Description
-- | -- | --
1.0.8802.1.1.2.1.3.8.1.1 | lldpLocManAddrSubtype | The type of management address identifier encoding used in the associated 'lldpLocManagmentAddr' object
1.0.8802.1.1.2.1.3.8.1.2 | lldpLocManAddr | The string value used to identify the management address component associated with the local system. The purpose of this address is to contact the management entity.
1.0.8802.1.1.2.1.3.8.1.3 | lldpLocManAddrLen | The total length of the management address subtype and the management address fields in LLDPDUs transmitted by the local LLDP agent.
1.0.8802.1.1.2.1.3.8.1.4 | lldpLocManAddrIfSubtype | The enumeration value that identifies the interface numbering method used for defining the interface number, associated with the local system
1.0.8802.1.1.2.1.3.8.1.5 | lldpLocManAddrIfId | The integer value used to identify the interface number regarding the management address component associated with the local system.
1.0.8802.1.1.2.1.3.8.1.6 | lldpLocManAddrOID | The OID value used to identify the type of hardware component or protocol entity associated with the management address advertised by the local system agent.


src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -37,8 +38,95 @@ class LLDPLocalChassis(int, Enum):
     lldp_loc_chassis_id = 2
     lldp_loc_sys_name = 3
     lldp_loc_sys_desc = 4
-    # *lldp_rem_sys_cap_supported = 5
-    # *lldp_rem_sys_cap_enabled = 6
+    lldp_loc_sys_cap_supported = 5
+    lldp_loc_sys_cap_enabled = 6
+
+
+class ManAddrConst(Enum):
```
qiluo-msft: >Enum [](start = 19, length = 4)

Why oid is a Enum member?

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -37,8 +38,95 @@ class LLDPLocalChassis(int, Enum):
...
+    advertised by the local system agent."
+    """
+    man_addr_oid = (1, 3, 6, 1, 2, 1, 2, 2, 1, 1)
+
+
+def poll_lldp_entry_updates(pubsub):
+    msg = pubsub.get_message()
+
+    if not msg:
+        return None, None, None
```
qiluo-msft: >None, None, None [](start = 15, length = 16)

ret = None, None, None
so you can reuse.

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -37,8 +38,95 @@ class LLDPLocalChassis(int, Enum):
...
+
+    lldp_entry = msg["channel"].split(b":")[-1].decode()
+    data = msg['data']
```
qiluo-msft: Handle many exception cases?

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -37,8 +38,95 @@ class LLDPLocalChassis(int, Enum):
...
+        try:
+            _table_name = bytes(getattr(table_name, 'name', table_name), 'utf-8')
+            return self.loc_chassis_data[_table_name]
+        except KeyError as e:
+            mibs.logger.warning(" 0 - b'LOC_CHASSIS' missing attribute '{}'.".format(e))
+            return None
+
+    def table_lookup_integer(self, table_name):
+        subtype_str = self.table_lookup(table_name)
+        return int(subtype_str) if subtype_str is not None else None
```
qiluo-msft: >None [](start = 64, length = 4)

None is not an int. Is there an safe fallback value?

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -109,34 +196,54 @@ def update_data(self):
             redis_client = self.db_conn.get_redis_client(self.db_conn.APPL_DB)
             db = self.db_conn.db_map[self.db_conn.APPL_DB]["db"]
             self.pubsub = redis_client.pubsub()
-            self.pubsub.psubscribe("__keyspace@{}__:{}".format(db, "LLDP_ENTRY_TABLE:*"))
+            self.pubsub.psubscribe("__keyspace@{}__:{}".format(db, mibs.lldp_entry_table(b'Ethernet*')))
```
qiluo-msft: >Ethernet [](start = 91, length = 8)

You add assumption of port name. I think original one is more accurate.

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -109,34 +196,54 @@ def update_data(self):
...
+            if b"set" in data:
+                self.update_interface_data(interface.encode('utf-8'))
 
-            # ignore management interface
-            if interface == "eth0":
-                continue
-            # get interface from interface name
-            if_index = port_util.get_index_from_str(interface)
+    def local_port_num(self, sub_id):
+        if len(sub_id) <= 0:
```
qiluo-msft: ><= [](start = 23, length = 2)

==

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -109,34 +196,54 @@ def update_data(self):
...
+            return None
+        return int(sub_id)
 
-            if if_index is None:
-                # interface name invalid, skip this entry
-                logger.warning("Invalid interface name in {} in APP_DB, skipping"
-                               .format(lldp_entry))
-                continue
+    def local_port_id(self, sub_id):
+        if len(sub_id) <= 0:
```
qiluo-msft: ><= [](start = 23, length = 2)

==

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -109,34 +196,54 @@ def update_data(self):
...
+        if_name = self.oid_name_map[sub_id]
+        if if_name not in self.loc_port_data:
+            # no LLDP data for this interface--we won't report the local interface
+            return None
+        return self.if_alias_map[if_name]
 
-            if b"set" in data:
-                self.update_interface_data(interface.encode('utf-8'))
+    def port_table_lookup(self, sub_id, table_name):
+        if len(sub_id) <= 0:
```
qiluo-msft: ><= [](start = 23, length = 2)

==

src/sonic_ax_impl/mibs/ieee802_1ab.py
mykolaf: The function is used to get ChassisIdSubtype, which is expected to be 4(mac address). I could hardcode it here, but i don't see the point.
Returning None here will result in END_OF_MIB, which is not that bad. If something is missing from LOC_CHASSIS table, table_lookup() has a log for that. 
What is the point of having fallback value for ChassisIdSubtype, if probably will also be missing the ChassisId itself? They are provided by the same flow. Looking up ChassisId will return the same None if misssing.

src/sonic_ax_impl/mibs/ieee802_1ab.py
mykolaf: I added it to avoid spam of log message about 'eth0' not being a valid interface name. In the end, i just changed dbsyncd to not submit LLDP_ENTRY:eth0 to DB, as it is never used anyway.

src/sonic_ax_impl/mibs/ieee802_1ab.py
mykolaf: @qiluo-msft do you insist on more changes or agree with my previous comment?

src/sonic_ax_impl/mibs/ieee802_1ab.py
qiluo-msft: You could remove 'Enum' and make code cleaner. Mixing different types of constants are not like an enum.

---
In reply to: [205701552](https://github.com/Azure/sonic-snmpagent/pull/72#discussion_r205701552) [](ancestors = 205701552)

src/sonic_ax_impl/mibs/ieee802_1ab.py
qiluo-msft: I see. It's better to check 'not eth0'.

---
In reply to: [205791584](https://github.com/Azure/sonic-snmpagent/pull/72#discussion_r205791584) [](ancestors = 205791584)

PR #87, opened 2018-09-21
Remove alias_map.json

stepanblyschak:
This file was added long time ago and now it is not needed anymore.
We can simply get port aliases from DB.

This also means that aliases will be now updated from DB once
a reinit timeout, instead of beeing static in alias_map.json

**- What I did**
I removed alias_map.json from SNMP agent. That map is not needed anymore since we can get port alias directly from DB

**- How I did it**
1. Removed /etc/snmp/alias_map.json parsing
2. Generate ```if_alias_map``` using aliases from APPL_DB
3. Added aliases in test appl_db.json table and fixed tests, that were expecting that alias == ifname

**- How to verify it**
Run unittests, run snmp agent on DUT and verify that we can still see correct aliases in SNMP MIBs

Remove alias_map.json

### DIFF
--- a/tests/test_sn.py
+++ b/tests/test_sn.py
@@ -66,7 +66,7 @@ class TestSonicMIB(TestCase):
         sub_id = 1000 * 1 # sub id for Ethernet100
 
         expected_mib = {
-            2: (ValueType.OCTET_STRING, "QSFP+ for Ethernet0"),
+            2: (ValueType.OCTET_STRING, "QSFP+ for etp1"),
             5: (ValueType.INTEGER, PhysicalClass.PORT),
             7: (ValueType.OCTET_STRING, ""), # skip
             8: (ValueType.OCTET_STRING, "A1"),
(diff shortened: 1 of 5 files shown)


tests/test_sn.py
```diff
@@ -97,20 +97,20 @@ def test_getpdu_xcvr_info(self):
...
-            1000 * 1 + 22: "DOM TX Bias Sensor for Ethernet0/2",
-            1000 * 1 + 32: "DOM TX Bias Sensor for Ethernet0/3",
-            1000 * 1 + 42: "DOM TX Bias Sensor for Ethernet0/4",
-            1000 * 1 + 13: "DOM TX Power Sensor for Ethernet0/1",
-            1000 * 1 + 23: "DOM TX Power Sensor for Ethernet0/2",
-            1000 * 1 + 33: "DOM TX Power Sensor for Ethernet0/3",
-            1000 * 1 + 43: "DOM TX Power Sensor for Ethernet0/4",
+            1000 * 1 + 1: "DOM Temperature Sensor for etp1",
+            1000 * 1 + 2: "DOM Voltage Sensor for etp1",
+            1000 * 1 + 11: "DOM RX Power Sensor for etp1/1",
```
lguohan: according to the naming convention it should be etp1a, etp1b, etp1c, etp1d

lguohan: https://github.com/Azure/SONiC/blob/master/doc/sonic-port-name.md

stepanblyschak: According to requirement doc for SNMP transceiver sensor monitoring the format  is "alias/lane_number".
Not all platforms currently have aliases in new naming convention, so the alias will be the name "EthernetN".
Also, even if we have a split port 'etp1a', the MIB table for 'etp1a' will contain sensor info for all 4 lanes, so it is necessary to specify lane number in description.
I can change this to "alias (lane X)" - that may be less confusing ?.

qiluo-msft:
Could you also remove unused file in buildimage repo?

PR #91, opened 2018-10-16
Fix LLDP local port description field

stepanblyschak:
Return empty string if port description is not present in DB

Management port doesn't have description in DB so return empty instead of returning None

**- What I did**

**- How I did it**

**- How to verify it**


qiluo-msft:
This fix could not be directly apply to 201803 branch.

PR #92, opened 2018-11-02
Allow alternative agentx socket configuration

MichelMoriniaux:
The point of this commit is to open up the snmp container to other agentx implementations in other containers.
By default the snmp agentx implementation uses '/var/agentx/master' unix domain socket as per RFC.
If we want other containers to be able to connect to the snmp container we need to expose this unix socket outside the container. To do this the unix socket needs to be instantiated on the host then exposed via a mount or volume to the containers that wish to use it including the snmp container. This is a good option which allows securing the socket via Unix file system permissions but in reality these permissions end up being too wide and a LOT of files need to be touched in a lot of repos to get this to work. additionally this breaks backwards compatibility. We believe there is a better way.

In this implementation we chose to use the snmp provisions to override the default socket via a configuration directive in the /etc/snmp/snmpd.conf file:
agentxsocket    tcp:localhost:705
we modify the ax_implementation so that this directive is read and the client connects to the defined tcp port, modifies unix socket or the default socket if the directive is not present. Thus we do not break behavior for existing configs and we allow the use of the facility in case it is configured.

The main real-world direct application of this is to allow the docker-fpm-frr container to feed bgp metrics to the snmp-container so that they can be polled.
Once the frr container is built with snmp support, the addition of a /etc/snmp/frr.conf with a single line containing "agenXSocket  tcp:localhost:705" does the trick and BGP information can be polled from the switch via the BGPv4 MIB.
 

**- What I did**

**- How I did it**

**- How to verify it**


src/ax_interface/socket_io.py
```diff
@@ -25,6 +26,20 @@ def __init__(self, mib_table, run_event, loop):
 
         self.transport = self.ax_socket = None
 
+        self.ax_socket_path = constants.AGENTX_SOCKET_PATH
+        # open the snmpd config file and search for a agentx socke redefinition
+        pattern = re.compile("^agentxsocket\s+(.*)$")
+        try :
+            with open(constants.SNMPD_CONFIG_FILE,'rt') as config_file:
+                for line in config_file:
+                    for match in pattern.finditer(line):
```
qiluo-msft: >for match in pattern.finditer(line): [](start = 20, length = 36)

why we need a loop if we are sure there is only none of one match?

src/ax_interface/socket_io.py
```diff
@@ -25,6 +26,20 @@ def __init__(self, mib_table, run_event, loop):
...
+                        self.ax_socket_path = match.group(1)
+        except:
+            log_level = logging.WARNING
+            logger.log(log_level, "SNMPD config file not found, using default agentx file socket")
+        log_level = logging.INFO
```
qiluo-msft: Add a blank line after 'try-except'.

src/ax_interface/socket_io.py
```diff
@@ -37,10 +52,21 @@ def __init__(self, mib_table, run_event, loop):
             try:
                 logger.info("Attempting AgentX socket bind...".format())
 
-                connection_routine = self.loop.create_unix_connection(
-                    protocol_factory=lambda: AgentX(self.mib_table, self.loop),
-                    path=constants.AGENTX_SOCKET_PATH,
-                    sock=self.ax_socket)
+                # Open the connection to the Agentx socket, we check the socket string to 
+                # either open a tcp socket or a Unix domain socket
+                if self.ax_socket_path.startswith('tcp'):
```
qiluo-msft: >if self.ax_socket_path.startswith('tcp'): [](start = 16, length = 41)

Is domain socket good enough? [http://www.net-snmp.org/docs/man/snmpd.html](http://www.net-snmp.org/docs/man/snmpd.html) shows many other protocols. We can report 'not implemented' for others we don't expect.

src/ax_interface/socket_io.py
```diff
@@ -25,6 +26,20 @@ def __init__(self, mib_table, run_event, loop):
...
+        try :
+            with open(constants.SNMPD_CONFIG_FILE,'rt') as config_file:
+                for line in config_file:
+                    for match in pattern.finditer(line):
+                        self.ax_socket_path = match.group(1)
+        except:
+            log_level = logging.WARNING
+            logger.log(log_level, "SNMPD config file not found, using default agentx file socket")
+        log_level = logging.INFO
+        logger.log(log_level, "Using agentx socket "+self.ax_socket_path)
```
qiluo-msft: >+ [](start = 52, length = 1)

add blanks before and after operator.

src/ax_interface/socket_io.py
```diff
@@ -25,6 +26,20 @@ def __init__(self, mib_table, run_event, loop):
 
         self.transport = self.ax_socket = None
 
+        self.ax_socket_path = constants.AGENTX_SOCKET_PATH
+        # open the snmpd config file and search for a agentx socke redefinition
+        pattern = re.compile("^agentxsocket\s+(.*)$")
```
qiluo-msft: >pattern [](start = 8, length = 7)

There is a corner case. 'snmpd -x' will override this config. How do you handle it?

src/ax_interface/socket_io.py
qiluo-msft: Can frr use domain socket?

---
In reply to: [234820946](https://github.com/Azure/sonic-snmpagent/pull/92#discussion_r234820946) [](ancestors = 234820946)

qiluo-msft:
Could you refine the PR title, with at least one verb.

qiluo-msft:
May I ask if vanilla quagga also has the capability? Not limited to FRR?

src/ax_interface/socket_io.py
MichelMoriniaux: frr can use a domain socket but that would involve mounting the domain socket from the host to all the containers. Instead of only modifying sonic-snmpagent we would also need to modify sonic-fpm-frr and sonic-buildimage to mount /var/agentx into the containers from the host.
This was actually my first implementation; but I found that going with the tcp port was way simpler and less intrusive in terms of modifications.

src/ax_interface/socket_io.py
MichelMoriniaux: true, I don't think there is a way to poll the daemon for autodiscovery of the AgentX address, we could use _psutils_ to look for the process and parse its 'cmdline' attribute. I think this is a corner case that should be documented that if you want to override the agentx socket use the config file.

MichelMoriniaux:
> May I ask if vanilla quagga also has the capability? Not limited to FRR?

I wasn't able to make it work on Quagga, even though documentation says it should work. I was not able to compile a working snmp version of Quagga.

MichelMoriniaux:
> Could you refine the PR title, with at least one verb?

Changed

src/ax_interface/socket_io.py
MichelMoriniaux: more genric IMHO, but corrected to one

src/ax_interface/socket_io.py
MichelMoriniaux: done

src/ax_interface/socket_io.py
```diff
@@ -25,6 +26,20 @@ def __init__(self, mib_table, run_event, loop):
...
+        try :
+            with open(constants.SNMPD_CONFIG_FILE,'rt') as config_file:
+                for line in config_file:
+                    for match in pattern.finditer(line):
+                        self.ax_socket_path = match.group(1)
+        except:
+            log_level = logging.WARNING
+            logger.log(log_level, "SNMPD config file not found, using default agentx file socket")
```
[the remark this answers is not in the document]
MichelMoriniaux: corrected

src/ax_interface/socket_io.py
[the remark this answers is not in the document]
MichelMoriniaux: modified, seems it is defined differently in different sources, made it case insensitive

MichelMoriniaux: Do you need me to implement the commandline parsing or can we live without it?
I will add a code comment that points to the corner case for future reference

qiluo-msft: No need if you add the code comment.

---
In reply to: [235185409](https://github.com/Azure/sonic-snmpagent/pull/92#discussion_r235185409) [](ancestors = 235185409)

src/ax_interface/socket_io.py
MichelMoriniaux: implemented methods for udp, unix and tcp, also the default of udp port if only a number is passed

qiluo-msft: I am ok with tcp socket. Still I think you don't need that much effort to use domain socket, and you can mount one folder in A docker to B docker.

---
In reply to: [234846409](https://github.com/Azure/sonic-snmpagent/pull/92#discussion_r234846409) [](ancestors = 234846409)

qiluo-msft: I am not sure I understand your statement. but I mean "We can report 'not implemented' for others we don't expect.", such as ipx, ssh, etc.

---
In reply to: [235197136](https://github.com/Azure/sonic-snmpagent/pull/92#discussion_r235197136) [](ancestors = 235197136)

MichelMoriniaux:
changes addressed and commits done

src/ax_interface/socket_io.py
```diff
@@ -37,10 +59,52 @@ def __init__(self, mib_table, run_event, loop):
             try:
                 logger.info("Attempting AgentX socket bind...".format())
 
-                connection_routine = self.loop.create_unix_connection(
-                    protocol_factory=lambda: AgentX(self.mib_table, self.loop),
-                    path=constants.AGENTX_SOCKET_PATH,
-                    sock=self.ax_socket)
+                # Open the connection to the Agentx socket, we check the socket string to 
+                # either open a tcp socket or a Unix domain socket
+                if '/' in self.ax_socket_path:
```
qiluo-msft: >if '/' in self.ax_socket_path: [](start = 16, length = 30)

As document ([http://www.net-snmp.org/docs/man/snmpd.html](http://www.net-snmp.org/docs/man/snmpd.html)) said, ipx may contains '/'. I think you will hit an exception.

MichelMoriniaux: yes and there are a bunch of other ones that I saw, I'm going to do a specific function to deal with it

qiluo-msft:
> Once the frr container is built with snmp support, the addition of a /etc/snmp/frr.conf with a single line containing "agenXSocket tcp:localhost:705" does the trick and BGP information can be polled from the switch via the BGPv4 MIB.

@MichelMoriniaux  Since FRR is the default routing stack on master branch, would you like continue the integration?

MichelMoriniaux:
I think we'll need to create a build option in rules/config as I believe
there is a licencing issue that prevents to distribute binary frr packages
with snmp support built-in. So we would need to put that option in so that
the images available on the Azure Jenkins don't get you in trouble.
What do you think?

pavel-shirshov:
Hi Michel,

Can you please elaborate on the licensing issues here?
Is there a special license for frr - snmp integration?

Thanks

MichelMoriniaux:
The following is extracted from: sonic-frr/frr/debianpkg/README.Debian

* Why has SNMP support been disabled?
=====================================
FRR used to link against the NetSNMP libraries to provide SNMP
support. Those libraries sadly link against the OpenSSL libraries
to provide crypto support for SNMPv3 among others.
OpenSSL now is not compatible with the GNU GENERAL PUBLIC LICENSE (GPL)
licence that FRR is distributed under. For more explanation read:
  http://www.gnome.org/~markmc/openssl-and-the-gpl.html
  http://www.gnu.org/licenses/gpl-faq.html#GPLIncompatibleLibs
Updating the licence to explecitly allow linking against OpenSSL
would requite the affirmation of all people that ever contributed
a significant part to Zebra / Quagga or FRR and thus are the collective
"copyright holder". That's too much work. Using a shrinked down
version of NetSNMP without OpenSSL or convincing the NetSNMP people
to change to GnuTLS are maybe good solutions but not reachable
during the last days before the Sarge release :-(

        *BUT*

It is allowed by the used licence mix that you fetch the sources and
build FRR yourself with SNMP with
        <remove the "grep ^smux" block at the end of debian/frr.preinst>
        # export WANT_SNMP=1
        # apt-get -b source frr
Just distributing it in binary form, linked against OpenSSL, is forbidden.


best regards,
Michel

pavel-shirshov:
Hi Michel,

Thank you for the valuable information.

Thanks

MichelMoriniaux:
I will start implementation on the rules/config option next week, that way
we will have choice

qiluo-msft:
Thanks @MichelMoriniaux!
I notice that https://github.com/FRRouting/frr/releases/tag/frr-6.0.2 isolated frr-snmp from frr. This is also the case of debian packages:
https://packages.debian.org/buster/frr
https://packages.debian.org/buster/frr-snmp

Seems a workable solution we could follow.

MichelMoriniaux:
Agreed, I will still implement the current solution as we have only
recently moved to FRR 4.0 and Stretch. so it might take some time before we
go to 6.0 and Buster

qiluo-msft:
Just to clarify, the master branch of sonic-buildimage is building frr 6.0.2-0 into a stretch docker image. I do see frr-snmp_6.0.2-0_amd64.deb in Jenkins workspace, but I have never tested it.

MichelMoriniaux:
oooh nice! I must've missed the last train. OK I'll add investigating all
that to my sprint next week and come up with the most standard non
disruptive way of doing it

pavel-shirshov:
Now we have frr-7.0 in sonic-master

MichelMoriniaux:
Shiny! do we have an idea of when we will be getting 6.0 or 7.0 in the
"stable" branches? I'm right now moving my org to 201904 but we are still
heavily invested in 201811

MichelMoriniaux:
please see sonic-buildimage PR 2981

PR #98, opened 2019-02-08
Enable overriding interface counters OIDs

qiluo-msft:
Depends on PR https://github.com/Azure/sonic-py-swsssdk/pull/48

This is a debug feature which enables network engineer (NE) override interface counters, in order to test network monitoring system. Sample scenarios which may be useful:
1. high link utilization
2. CRC error on links
3. imbalance link utilization

The override OID values are stored in a new Redis database. SNMP subagent will fetch them frequently and override existing ones if OID matches. Otherwise:
1. If existing ones has no match in Redis, then keep existing one
2. If Redis has extra OID than existing one, they will be ignored

In concept, this is a overlay OID subtree
1. overlay layer is Redis subtree
2. underlay layer is existing subtree

Usage:
```
root@str-msn2700-03:/# snmpwalk -v2c -c public 127.0.0.1 1.3.6.1.2.1.2.2.1.10.1
iso.3.6.1.2.1.2.2.1.10.1 = Counter32: 0

admin@str-msn2700-03:~$ redis-cli -n 7 hmset "1.3.6.1.2.1.2.2.1.10.1" "type" "COUNTER_32" "data" "54321"
OK

root@str-msn2700-03:/# snmpwalk -v2c -c public 127.0.0.1 1.3.6.1.2.1.2.2.1.10.1
iso.3.6.1.2.1.2.2.1.10.1 = Counter32: 54321

admin@str-msn2700-03:~$ redis-cli -n 7 del "1.3.6.1.2.1.2.2.1.10.1"
(integer) 1

root@str-msn2700-03:/# snmpwalk -v2c -c public 127.0.0.1 1.3.6.1.2.1.2.2.1.10.1
iso.3.6.1.2.1.2.2.1.10.1 = Counter32: 0
```

Sample script to emulate a link with hot inbound traffic
```python
#!/usr/bin/env python3
import time
import redis
import ctypes

r = redis.StrictRedis(db=7)

counter = 0
oid = "1.3.6.1.2.1.31.1.1.1.6.5"

try:
    r.hmset(oid, { "type": "COUNTER_64", "data": str(counter) })

    # Override interface counter every 5 seconds for 1 hour
    intv = 5
    for i in xrange(0, 3600 / intv):
        counter += 4500000000
        counter = ctypes.c_uint64(counter).value
        r.hset(oid, "data", str(counter))
        print("hset data %d" % counter)
        time.sleep(intv)
finally:
    # Clean up the overriding and use true counter
    r.delete(oid)
```


src/sonic_ax_impl/mibs/ietf/rfc1213.py
```diff
@@ -467,52 +468,88 @@ class InterfacesMIB(metaclass=MIBMeta, prefix='.1.3.6.1.2.1.2'):
...
+            SubtreeMIBEntry('2.1.10', if_updater, ValueType.COUNTER_32, if_updater.get_counter,
+                           DbTables(10)),
+            OidMIBEntry('2.1.10', ValueType.COUNTER_32, oidtree_updater.get_oidvalue)
+        )
 
     ifInUcastPkts = \
-        SubtreeMIBEntry('2.1.11', if_updater, ValueType.COUNTER_32, if_updater.get_counter,
-                           DbTables(11))
+        OverlayAdpaterMIBEntry(
+            SubtreeMIBEntry('2.1.11', if_updater, ValueType.COUNTER_32, if_updater.get_counter,
```
lguohan: why not just change the implementation of if_updater.get_counter to use the overlay db?

qiluo-msft: That will work in theory. The solution here has some benefit:
1. easy to understand DB schema, user don't need to understand port SAI object ID, SAI attribute, etc.
2. general and extensible to other future overriding case. we don't need overlay every database in long run. For example, one for counter_db, one for appl_db, etc.


---
In reply to: [255416371](https://github.com/Azure/sonic-snmpagent/pull/98#discussion_r255416371) [](ancestors = 255416371)

hui-ma: @qiluo-msft 

1. Who does the "user" here refer to? In the currently plan, is it going to be an agent on on sonic to write all the  overlay DB?

 2. If I understand it correctly, overlay DB here are based on snmp tree 1.3.6.1.2.1.2.2.1.10.1 OID and their value. Why does it need SAP object info?

qiluo-msft: 1. The user means anyone or any script that wants to override SNMP oid value
2. This PR is based on SNMP subtree override. There is another design choice that override counter DB directly, that one will need SAI object ID.
Let me know if you have more questions.

---
In reply to: [255627438](https://github.com/Azure/sonic-snmpagent/pull/98#discussion_r255627438) [](ancestors = 255627438)

qiluo-msft:
Retest this please

qiluo-msft:
Retest this please

PR #103, opened 2019-04-11
integrating py.test into setuptools

samaity:
Signed-off-by: Sangita Maity samaity@linkedin.com

- **What I did**
Integrate py.test into setuptools as vanilla 'python3 setup.py test' does not work.
This PR is needed for snmp stretch upgrade docker PR.

1. [https://github.com/Azure/sonic-buildimage/pull/2620](https://github.com/Azure/sonic-buildimage/pull/2620)

- **How to verify it**
Use this PR with snmp docker upgrade PR in sonic-buildimage.

Ref: https://docs.pytest.org/en/latest/goodpractices.html

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,11 +19,10 @@ setup(
     name='asyncsnmp',
     install_requires=dependencies,
     setup_requires=["pytest-runner"],
-    tests_require=["pytest"],
+    tests_require=test_deps,
     version='2.1.0',
     packages=find_packages('src'),
     extras_require={
-        'testing': test_deps,
         'high_perf': high_performance_deps,
     },
     license='Apache 2.0',
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,8 @@ high_performance_deps = [
 setup(
     name='asyncsnmp',
     install_requires=dependencies,
+    setup_requires=["pytest-runner"],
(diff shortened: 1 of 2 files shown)


samaity:
@qiluo-msft, kindly review it. We need it for snmp docker upgrade to stretch.

qiluo-msft:
>         'testing': test_deps,

Can we remove this line?

---
Refers to: setup.py:26 in 33346ee. [](commit_id = 33346ee1d3923c57081f5848452fdd22f99e14db, deletion_comment = False)

setup.py
```diff
@@ -18,6 +18,8 @@
 setup(
     name='asyncsnmp',
     install_requires=dependencies,
+    setup_requires=["pytest-runner"],
```
qiluo-msft: >setup_requires [](start = 4, length = 14)

Add pytest-runner into test_deps, and use test_deps here?

setup.py
```diff
@@ -18,6 +18,8 @@
...
+    tests_require=["pytest"],
```
qiluo-msft: >tests_require [](start = 4, length = 13)

Add pytest into test_deps, and use test_deps here?

samaity: resolved

PR #113, opened 2019-09-12
[LLDP MIB] fix lldp loc mgmt ip logic

mykolaf:
Signed-off-by: Mykola Faryma <mykolaf@mellanox.com>

**- What I did**
Fixed an issue, when the lldp_loc_mgmt_ip was unavailable in APP DB (at startup for example),
the reinit_data resulted in an ERR log:
```
ERR snmp#snmp-subagent [sonic_ax_impl] ERROR: Invalid local mgmt IP
```

With the fix, we will warn the user and return. In most cases the data will soon appear in DB. In other cases we get a warning every reinit_data(1m).

**- How I did it**

**- How to verify it**
stop lldp, snmp services
Flush LLDP_LOC_CHASSIS from  APPDB,
start snmp service

### DIFF
--- a/src/sonic_ax_impl/mibs/ieee802_1ab.py
+++ b/src/sonic_ax_impl/mibs/ieee802_1ab.py
@@ -296,6 +296,7 @@ class LLDPLocManAddrUpdater(MIBUpdater):
         Subclass update data routine.
         """
         self.man_addr_list = []
+        self.mgmt_ip_str = None
 
         # establish connection to application database.
         self.db_conn.connect(mibs.APPL_DB)
--- a/src/sonic_ax_impl/mibs/ieee802_1ab.py
+++ b/src/sonic_ax_impl/mibs/ieee802_1ab.py
@@ -302,10 +302,11 @@ class LLDPLocManAddrUpdater(MIBUpdater):
         mgmt_ip_bytes = self.db_conn.get(mibs.APPL_DB, mibs.LOC_CHASSIS_TABLE, b'lldp_loc_man_addr')
 
         if not mgmt_ip_bytes:
-            self.mgmt_ip_str = ''
-        else:
-            self.mgmt_ip_str = mgmt_ip_bytes.decode()
-            logger.debug("Got mgmt ip from db : {}".format(self.mgmt_ip_str))
+            logger.warning("Missing lldp_loc_man_addr from APPL DB")
+            return
+
+        self.mgmt_ip_str = mgmt_ip_bytes.decode()
+        logger.debug("Got mgmt ip from db : {}".format(self.mgmt_ip_str))
         try:
             addr_subtype_sub_oid = 4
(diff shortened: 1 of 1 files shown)


src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -302,10 +302,11 @@ def reinit_data(self):
         mgmt_ip_bytes = self.db_conn.get(mibs.APPL_DB, mibs.LOC_CHASSIS_TABLE, b'lldp_loc_man_addr')
 
         if not mgmt_ip_bytes:
-            self.mgmt_ip_str = ''
-        else:
-            self.mgmt_ip_str = mgmt_ip_bytes.decode()
-            logger.debug("Got mgmt ip from db : {}".format(self.mgmt_ip_str))
+            logger.warning("Missing lldp_loc_man_addr from APPL DB")
+            return
```
qiluo-msft: >return [](start = 12, length = 6)

if return here, self.mgmt_ip_str will be None or outdated value. Later updater.man_addr will throw or get outdated value.

mykolaf: I didn't observe any unhandled exceptions or logs like "Unexpecteed exception ..." while querrying for absent field in the DB. But anyway, **added a check to avoid throws caused by self.mgmt_ip = None.** 

Manually verified performing snmpwalk for unavailable OID we get  'No Such Object available on this agent at this OID', then pushed the data into DB and the OID became available. No throws/tracebacks/suspicious logs observed.

qiluo-msft: I see you fixed the issue 'self.mgmt_ip_str will be None'. However, The case 'outdated value' still there. The timeline is like:
1. AppDB has the correct lldp_loc_man_addr, this MIB works well
2. For whatever reason, this data disappear in AppDB
3. reinit_data() finds mgmt_ip_bytes == None and return
4. self.mgmt_ip_str is outdated. So MIB exposes wrong values.

---
In reply to: [325766888](https://github.com/Azure/sonic-snmpagent/pull/113#discussion_r325766888) [](ancestors = 325766888)

mykolaf: @qiluo-msft 
In the described timeline the MIB will not expose wrong/outdated value. What happens is on reinit_data(), the updater [drops the self.man_addr_list](https://github.com/Azure/sonic-snmpagent/blob/5533752301c95eb483a609eeccd1f148c90944a2/src/sonic_ax_impl/mibs/ieee802_1ab.py#L298), so the updater will return None on get_next(). I could not get the outdated value via snmpwalk niether by querrying the full MIB, nor specific OID. After removing data from DB, the reinit picks up the change and the OID becomes unavailable.

I understand your concern, let's drop the mgmt_ip_string on reinit too. Just to be sure.

PR #116, opened 2019-10-08
Using get API to obtain dbid instead of internal var directly

dzhangalibaba:
* application should use get api to obtain dbid instead of accessing internal variable  

* later for multi DB changes, we will redesign the internal db_map variable and implementation of get_dbid API , so to avoid breakage, we should use API to obtain dbid and merge snmpagent first.

Signed-off-by: Dong Zhang d.zhang@alibaba-inc.com


dzhangalibaba:
@qiluo-msft 

1. application  should use get api to obtain dbid instead of accessing internal variable  

2. later for multi DB changes, we will remove the internal db_map variable and redesign the implementation of get_dbid() API , so to avoid breakage, we should use get API to obtain dbid and merge this snmpagent changes first.

PR #119, opened 2019-12-09
Implement cbgpPeer2State in CiscoBgp4MIB

qiluo-msft:
The implementation is using a TCP socket talking with Quagga/FRR daemon


src/sonic_ax_impl/lib/perseverantsocket.py
```diff
@@ -0,0 +1,35 @@
+import socket
+
+class PerseverantSocket:
+    def __init__(self, family=socket.AF_INET, type=socket.SOCK_STREAM, proto=0, fileno=None, address_tuple=None, *args, **kwargs):
+        self._connected = False
+        self.address_tuple = address_tuple
+        self.family = family
+        self.type = type
```
pavel-shirshov: type is already used by a standard python library.

qiluo-msft: I am following parameter naming in [https://docs.python.org/3/library/socket.html#socket.socket](https://docs.python.org/3/library/socket.html#socket.socket). Since it is safe, let's keep it.

---
In reply to: [355809292](https://github.com/Azure/sonic-snmpagent/pull/119#discussion_r355809292) [](ancestors = 355809292)

src/sonic_ax_impl/lib/quaggaclient.py
```diff
@@ -90,6 +90,7 @@ class QuaggaClient:
     def __init__(self, hostname, sock):
         self.prompt_hostname = ('\r\n' + hostname + '> ').encode()
         self.sock = sock
+        self.bpg_provider = 'Quagga'
```
pavel-shirshov: bgp_provider?

src/sonic_ax_impl/lib/quaggaclient.py
```diff
@@ -99,16 +100,27 @@ def union_bgp_sessions(self):
...
+
+        ## Nowadays we see 2 BGP stack
+        ## 1. Quagga (version 0.99.24.1)
+        ## 2. FRRouting (version 7.2-sonic)
+        banner = self.vtysh_recv()
+        if 'Quagga' in banner:
+            self.bpg_provider = 'Quagga'
+        elif 'FRRouting' in banner:
+            self.bpg_provider = 'FRRouting'
+        else:
```
pavel-shirshov: I'd suggest to use json output in case of FRR. FRR supports it and also FRR output could be changed,

src/sonic_ax_impl/lib/quaggaclient.py
```diff
@@ -0,0 +1,150 @@
...
+def parse_bgp_summary(summ):
+    ls = summ.splitlines()
+    bgpinfo = []
+
+    ## Read until the table header
+    n = len(ls)
+    li = 0
+    while li < n:
+        l = ls[li]
+        if l.startswith('Neighbor        '): break
```
pavel-shirshov: break on the next line?

src/sonic_ax_impl/lib/quaggaclient.py
qiluo-msft: Will do in future PR.

---
In reply to: [356096823](https://github.com/Azure/sonic-snmpagent/pull/119#discussion_r356096823) [](ancestors = 356096823)

src/sonic_ax_impl/lib/quaggaclient.py
```diff
@@ -0,0 +1,151 @@
...
+            break
+        if l.startswith('No IPv'): # eg. No IPv6 neighbor is configured
+            return bgpinfo
+        if l.endswith('> exit'): # last command in the lines
+            return bgpinfo
+        li += 1
+
+    ## Read and store the table header
+    if li >= n:
+        raise ValueError('No table header found')
```
yxieca: This error message would be clearer if it says "no neighbor information found".

PR #120, opened 2019-12-12
Implement cbgpPeer2State in CiscoBgp4MIB (#119)

qiluo-msft:
The same feature as https://github.com/Azure/sonic-snmpagent/pull/119 but for 201811 branch.


qiluo-msft:
The test failure is because the `dep-build` does not provide 201811 binaries. I tested it locally.

PR #122, opened 2020-01-04
Fix Quagga/FRR parser on IPv6 BGP sessions

qiluo-msft:
If there is no ipv4 or ipv6 BGP neighbors configured, the messages are a little different in different Quagga/FRR version.

Quagga in 201811
```
sonic> show ipv6 bgp su
No IPv6 neighbor is configured
sonic> 
```
FRR in 201811

```
sonic> show ip bgp ipv6 summary
sonic>
```

FRR in master

```
sonic> show ip bgp ipv6 summary
% No BGP neighbors found
sonic> 
```
 Currently we could not handle 2nd or 3rd cases.

### DIFF
--- a/src/sonic_ax_impl/lib/quaggaclient.py
+++ b/src/sonic_ax_impl/lib/quaggaclient.py
@@ -22,15 +22,17 @@ def parse_bgp_summary(summ):
         l = ls[li]
         if l.startswith('Neighbor        '):
             break
-        if l.startswith('No IPv'): # eg. No IPv6 neighbor is configured
+        if l.startswith('No IPv'): # eg. No IPv6 neighbor is configured, in Quagga (version 0.99.24.1)
             return bgpinfo
-        if l.endswith('> exit'): # last command in the lines
+        if l.startswith('% No BGP neighbors found'): # in FRRouting (version 7.2)
+            return bgpinfo
+        if l.endswith('> '): # directly hostname prompt, in FRRouting (version 4.0)
             return bgpinfo
         li += 1
 
     ## Read and store the table header
     if li >= n:
-        raise ValueError('No table header found')
+        raise ValueError('No table header found: ' + summ)
     hl = ls[li]
     li += 1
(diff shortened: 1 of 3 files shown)


qiluo-msft:
@zhenggen-xu Could you please help review and test in your DUT?

src/sonic_ax_impl/lib/quaggaclient.py
```diff
@@ -22,15 +22,17 @@ def parse_bgp_summary(summ):
         l = ls[li]
         if l.startswith('Neighbor        '):
             break
-        if l.startswith('No IPv'): # eg. No IPv6 neighbor is configured
```
zhenggen-xu: Can we define these strings as constant at beginning so it would be easier to maintain?

zhenggen-xu:
> @zhenggen-xu Could you please help review and test in your DUT?

Test worked fine with 201811 branch.

src/sonic_ax_impl/lib/quaggaclient.py
yxieca: I think this is a great idea. We could define a list of criteria to skip (with comments) at one place and check them in a loop here.

qiluo-msft: In my opinion, both the strings and the functions such as startswith or endswith are important. So keep them together help maintenance.

---
In reply to: [363017378](https://github.com/Azure/sonic-snmpagent/pull/122#discussion_r363017378) [](ancestors = 363017378)

yxieca: You could define 2 lists of strings. one for starts with and one for ends with. They are still defined together. And the code to make use of these 2 lists are also together. This way, the data and handling are separate. I think it is cleaner this way.

qiluo-msft: We can improve this in future. This is a quick bug fix for 201811 FRR special use case required by @zhenggen-xu . Since it is tested by him, let's keep the code in this PR.

PR #124, opened 2020-03-07
Parse quagga output without knowledge about hostname, so robust against hostname changes or mismatch

qiluo-msft:


src/sonic_ax_impl/lib/quaggaclient.py
```diff
@@ -133,11 +134,22 @@ def vtysh_run(self, command):
...
-            if acc.endswith(self.prompt_hostname):
+            ## 1. To match hostname
+            ##    RFC 1123 Section 2.1
+            ##    Hostname starts with with alphabet or number
+            ##    Hostname lenght <= 255
+            ##    Hostname contains no whitespace characters
+            ## 2. To match the prompt line
+            ##    The buffer may containers only prompt without return char
+            ##    Or the buffer container some output followed by return char and prompt
+            if re.search(b'(^|\r\n)\\w[\\S]{0,254}[#>] $', acc):
```
lguohan: the regex looks good to me

src/sonic_ax_impl/lib/quaggaclient.py
```diff
@@ -133,11 +134,22 @@ def vtysh_run(self, command):
...
+                raise ValueError('Timeout recv acc=: {0}'.format(acc)) from e
             if not data:
-                raise ValueError('Unexpected data recv: {0}'.format(acc))
+                raise ValueError('Unexpected data recv acc=: {0}'.format(acc))
             acc += data
-            if acc.endswith(self.prompt_hostname):
+            ## 1. To match hostname
+            ##    RFC 1123 Section 2.1
+            ##    Hostname starts with with alphabet or number
+            ##    Hostname lenght <= 255
```
jleveque: s/lenght/length/

src/sonic_ax_impl/lib/quaggaclient.py
jleveque: While it should never be a problem in this use case, technically, the `\\w` to match the first char of the host name is slightly too lax, as it includes `_`. A more strict regex would be `'(^|\r\n)[a-zA-Z0-9][\\S]{0,254}[#>] $' `.

src/sonic_ax_impl/lib/quaggaclient.py
```diff
@@ -133,11 +134,22 @@ def vtysh_run(self, command):
...
+            except socket.timeout as e:
+                raise ValueError('Timeout recv acc=: {0}'.format(acc)) from e
             if not data:
-                raise ValueError('Unexpected data recv: {0}'.format(acc))
+                raise ValueError('Unexpected data recv acc=: {0}'.format(acc))
             acc += data
-            if acc.endswith(self.prompt_hostname):
+            ## 1. To match hostname
+            ##    RFC 1123 Section 2.1
+            ##    Hostname starts with with alphabet or number
```
jleveque: Suggest to change to `First char of hostname must be letter or number`

src/sonic_ax_impl/lib/quaggaclient.py
```diff
@@ -88,10 +89,9 @@ def bgp_peer_tuple(dic):
 class QuaggaClient:
     HOST = '127.0.0.1'
     PORT = 2605
-    PROMPT_PASSWORD = b'Password: '
+    PROMPT_PASSWORD = b'\x1fPassword: '
```
jleveque: Can you please briefly explain the need to add the `\x1f` byte prefix?

src/sonic_ax_impl/lib/quaggaclient.py
qiluo-msft: Resolve, I even change more words

---
In reply to: [389239624](https://github.com/Azure/sonic-snmpagent/pull/124#discussion_r389239624) [](ancestors = 389239624)

src/sonic_ax_impl/lib/quaggaclient.py
qiluo-msft: I agree your proposal conforms to RFC, but I purposely make it relax this way because FRR implementation has `hostname [NAME]` command and did input validation as below
```c++
 if (!isalnum((unsigned char)word->arg[0])) {
  vty_out(vty,
      "Please specify string starting with alphabet or number\n");
  return CMD_WARNING_CONFIG_FAILED;
 }

 /* With reference to RFC 1123 Section 2.1 */
 if (strlen(word->arg) > HOSTNAME_LEN) {
  vty_out(vty, "Hostname length should be less than %d chars\n",
   HOSTNAME_LEN);
  return CMD_WARNING_CONFIG_FAILED;
 }
```

---
In reply to: [389239562](https://github.com/Azure/sonic-snmpagent/pull/124#discussion_r389239562) [](ancestors = 389239562)

src/sonic_ax_impl/lib/quaggaclient.py
qiluo-msft: It's observation of Quagga and FRR implementations.

---
In reply to: [389239735](https://github.com/Azure/sonic-snmpagent/pull/124#discussion_r389239735) [](ancestors = 389239735)

src/sonic_ax_impl/lib/quaggaclient.py
jleveque: I understand, and as I said previously, in this use case it shouldn't be an issue. However, it also wouldn't hurt to make the regex more precise (it might just make it a bit uglier).

src/sonic_ax_impl/lib/quaggaclient.py
jleveque: Just curious why it wasn't necessary before?

src/sonic_ax_impl/lib/quaggaclient.py
qiluo-msft: resolved

---
In reply to: [389307576](https://github.com/Azure/sonic-snmpagent/pull/124#discussion_r389307576) [](ancestors = 389307576)

src/sonic_ax_impl/lib/quaggaclient.py
qiluo-msft: It is not a bug fix, but an improvement in the parser.

---
In reply to: [389307626](https://github.com/Azure/sonic-snmpagent/pull/124#discussion_r389307626) [](ancestors = 389307626)

abdosi:
@qiluo-msft Don we need this for 201911 ? As we are using FRR in 209111

qiluo-msft:
@abdosi Yes for 201911

PR #131, opened 2020-05-27
[Multi-asic]: Namespace support for LLDP and Sensor tables

SuvarnaMeenakshi:
Signed-off-by: SuvarnaMeenakshi <sumeenak@microsoft.com>

**- What I did**
This PR targets changes in lldp/sensor related MIB implementation for multi-asic platform.
LLDP Mgmt interface related information will be obtained from global db.
Update implementation of the following MIBS to get data from all
databases:
    LLDPRemManAddrTable
    PhysicalTableMIB
    PhysicalSensorTableMIB
    QBridgeMIBObjects
**- How I did it**
Modified implementation of above MIB classes to obtain data from all namespace database DBs.

**- How to verify it**
Verified in multi-asic platform.
Note: Requires code/test mock tables from https://github.com/Azure/sonic-snmpagent/pull/126 

**- Description for the changelog**

PR #134, opened 2020-06-16
Fix undefined variable and warning message

SuvarnaMeenakshi:
**- What I did**
- Fix undefined variable introduced in https://github.com/Azure/sonic-snmpagent/pull/131
- Add fdb unit-test for namespace changes
- Fix update_data()  LocPortUpdater, LLDPRemManAddrUpdater and PhysicalTableMIBUpdater caused due to Namespace changes. The pubsub channel should not be created everytime on update-data and should be created only if it is not previously created. This causes a warning message and does not get any new changes in the subscribed key pattern.
```
sonic WARNING snmp#snmp-subagent [sonic_ax_impl] WARNING: Invalid interface name in TRANSCEIVER_INFO|*                      in STATE_DB, skipping
snmp#snmp-subagent [sonic_ax_impl] WARNING: Invalid interface name in *' in APP_DB, skipping
```

**- How I did it**
- Added fdb unit-test for namespace changes
- Made sure that the pubsub channel for each Namespace DB is created only once.

**- How to verify it**
- Tested out on single asic and multiple asic platforms.
Make sure that warning messages are not seen in syslog.


src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -536,18 +544,18 @@ def update_rem_if_mgmt(self, if_oid, if_name):
             return
         self.if_range.sort()
 
-    def _update_per_namespace_data(self, db_conn, pubsub):
```
qiluo-msft, line 539: >_update_per_namespace_data [](start = 8, length = 26)

Original code seems better. Why change?

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -190,14 +192,14 @@ def _update_per_namespace_data(self, statedb, pubsub):
...
-        if not pubsub:
-            redis_client = statedb.get_redis_client(statedb.STATE_DB)
-            db = statedb.get_dbid(statedb.STATE_DB)
-            pubsub = redis_client.pubsub()
-            pubsub.psubscribe("__keyspace@{}__:{}".format(db, self.TRANSCEIVER_KEY_PATTERN))
+        if not self.pubsub[inst]:
+            redis_client = self.statedb[inst].get_redis_client(self.statedb[inst].STATE_DB)
+            db = self.statedb[inst].get_dbid(self.statedb[inst].STATE_DB)
+            self.pubsub[inst] = redis_client.pubsub()
+            self.pubsub[inst].psubscribe("__keyspace@{}__:{}".format(db, self.TRANSCEIVER_KEY_PATTERN))
```
qiluo-msft, line 199: Duplicated code. Could you reuse?

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -104,6 +104,7 @@ def __init__(self):
         super().__init__()
 
         self.db_conn = Namespace.init_namespace_dbs()
+        Namespace.connect_all_dbs(self.db_conn, mibs.APPL_DB)
```
qiluo-msft, line 107: >Namespace.connect_all_dbs(self.db_conn, mibs.APPL_DB) [](start = 8, length = 53)

Original code could tolerate temporary or initial Redis not offline and recover whenever online. But the new code could not. Can you improve and test?

tests/mock_tables/asic0/asic_db.json
```diff
@@ -1,2 +1,27 @@
...
+    "SAI_BRIDGE_PORT_ATTR_TYPE": "SAI_BRIDGE_PORT_TYPE_PORT",
+    "SAI_BRIDGE_PORT_ATTR_PORT_ID": "oid:0x1000000000003",
+    "SAI_BRIDGE_PORT_ATTR_ADMIN_STATE": "true"
+  },
+  "ASIC_STATE:SAI_OBJECT_TYPE_BRIDGE_PORT:oid:0x3a000000000616": {
+    "SAI_BRIDGE_PORT_ATTR_TYPE": "SAI_BRIDGE_PORT_TYPE_PORT",
+    "SAI_BRIDGE_PORT_ATTR_PORT_ID": "oid:0x1000000000005",
+    "SAI_BRIDGE_PORT_ATTR_ADMIN_STATE": "true"
+  }
 }
```
qiluo-msft, line 27: If not related to the PR subject, could you submit another PR just for test cases?

src/sonic_ax_impl/mibs/ieee802_1ab.py
SuvarnaMeenakshi: Before namespace changes, self.pubsub was getting updated in update_data(). After namespace change, self.pubsub array was not updated as we pass the instance. By passing the array index, we can ensure that the array is directly getting udpated. So this fix is to make sure the code is as it was before namespace changes.

tests/mock_tables/asic0/asic_db.json
SuvarnaMeenakshi: This test case is to test the code path where the undefined variable is fixed in this PR.

src/sonic_ax_impl/mibs/ietf/rfc2737.py
SuvarnaMeenakshi: This similar pubsub update is done only 2 instances with different DB Name and pattern.

src/sonic_ax_impl/mibs/ieee802_1ab.py
SuvarnaMeenakshi: I do see that if redis-server process is stopped and started, the MIB walk does not get affected, as the connect is persistent_connect. But, I have reverted this change, I will raise a new PR to provide a clean fix to reduce runtime.

src/sonic_ax_impl/mibs/ieee802_1ab.py
SuvarnaMeenakshi: Updated description with the issue and why this fix is required.

src/sonic_ax_impl/mibs/ietf/rfc2737.py
qiluo-msft: Duplicated with ieee802_1ab.py

---
In reply to: [441920582](https://github.com/Azure/sonic-snmpagent/pull/134#discussion_r441920582) [](ancestors = 441920582)

src/sonic_ax_impl/mibs/ieee802_1ab.py
qiluo-msft: I see the bug in [https://github.com/Azure/sonic-snmpagent/pull/131](https://github.com/Azure/sonic-snmpagent/pull/131).

To make code cleaner, could you try add
```
db_conn = self.db_conn[inst]
pubsub = self.pubsub[inst]
if not pubsub:
    ...
    self.pubsub[inst] = pubsub

```

---
In reply to: [441919551](https://github.com/Azure/sonic-snmpagent/pull/134#discussion_r441919551) [](ancestors = 441919551)

SuvarnaMeenakshi: Fixed as per comments.

src/sonic_ax_impl/mibs/ietf/rfc2737.py
SuvarnaMeenakshi: Fixed as per review comments.

PR #137, opened 2020-06-30
[Namespace]: Reduce time taken during get_all from all namespaces

SuvarnaMeenakshi:
**- What I did**
To support namespaces, we added a Namespace class with functions to support getting data from all databases.
The Namespace dbs_get_all function , checks if the key exists in the database before executing get_all.
This additional check was causing additional time, which was causing agentx socket to disconnect.

**- How I did it**
- Added new Namespace class variables to get the mapping of interface index and  namespace(or database instance).
- Use the new mapping to directly execute get_all from that specific database instance instead of checking if the key exists in the database.

**- How to verify it**
Make sure unit-test passes.
Make sure single-asic platform test cases are passing and no error message is observed in the syslog.
Check snmp output in multi-asic platform.


src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -513,49 +532,55 @@ def init_namespace_sync_d_interface_tables(dbs):
...
+        if_oid_namespace = {}
 
         """
         all_ns_db - will have db_conn to all namespace DBs and
         global db. First db in the list is global db.
         Ignore first global db to get interface tables if there
         are multiple namespaces.
         """
-        for db_conn in Namespace.get_non_host_dbs(dbs):
+        for inst in range(Namespace.get_start_idx_of_non_host(dbs), len(dbs)):
```
qiluo-msft, line 543: >inst [](start = 12, length = 4)

ns_index or ns ?

SuvarnaMeenakshi: ns_index; basically, the index in db array.
index is not same as namespace, index 0 will be host namespace.

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -166,26 +172,31 @@ def reinit_data(self):
...
         for if_oid, if_name in self.oid_name_map.items():
-            self.update_interface_data(if_name)
+            self.update_interface_data(if_name, self.if_oid_namespace[if_oid])
             self.if_range.append((if_oid, ))
         self.if_range.sort()
         if not self.loc_port_data:
             logger.warning("0 - b'PORT_TABLE' is empty. No local port information could be retrieved.")
 
-    def _get_if_entry(self, if_name):
+    def _get_if_entry(self, if_name, db_inst):
```
qiluo-msft, line 199: >db_inst [](start = 37, length = 7)

db_index?

src/sonic_ax_impl/mibs/__init__.py
qiluo-msft: Then call it `db_index` ?

---
In reply to: [449314977](https://github.com/Azure/sonic-snmpagent/pull/137#discussion_r449314977) [](ancestors = 449314977)

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -504,7 +510,20 @@ def get_non_host_dbs(dbs):
...
+        For multiple namespace dbs, return all dbs except the
+        host namespace db which is the first db in the list.
+        """
+        if len(dbs) == 1:
+            return 0
+        else:
+            return 1
 
     @staticmethod
     def init_namespace_sync_d_interface_tables(dbs):
```
qiluo-msft, line 529: >init_namespace_sync_d [](start = 8, length = 21)

I see a common pattern in all the functions like `init_namespace_sync_d*`. They just call the single db function and merge all returned items respectively.

How about abstract this high level function and return a dictionary of list, in  this case the type is `db -> { if_name_map, if_alias_map, if_id_map, oid_sai_map, oid_name_map, if_oid_namespace }`

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -504,7 +510,20 @@ def get_non_host_dbs(dbs):
             return dbs
         else:
             return dbs[1:]
-        
+
+    @staticmethod
+    def get_start_idx_of_non_host(dbs):
```
qiluo-msft, line 515: >get_start_idx_of_non_host [](start = 8, length = 25)

You can return a range since all use cases are in a range. You can further implement get_non_host_dbs by this function.

A better name will be `get_non_host_db_indexes`

src/sonic_ax_impl/mibs/__init__.py
SuvarnaMeenakshi: In the MIB implementations, we generally use a single dict to look up the if_index etc. If we have a dict of lists, then we will have to loop through this dict for each instance of the db and look for the if_index, which will make the code look bulky in the places where these functions are used.

src/sonic_ax_impl/mibs/__init__.py

src/sonic_ax_impl/mibs/ieee802_1ab.py
SuvarnaMeenakshi: updated as per comment

src/sonic_ax_impl/mibs/__init__.py
SuvarnaMeenakshi: updated as per comment

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -233,9 +233,9 @@ def update_data(self):
             if not self.pubsub[i]:
                 pattern = self.TRANSCEIVER_KEY_PATTERN
                 self.pubsub[i] = mibs.get_redis_pubsub(self.statedb[i], self.statedb[i].STATE_DB, pattern)
```
qiluo-msft, line 235: Move to `__init__()` ?

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -168,20 +168,20 @@ def reinit_data(self):
...
         # update cache with initial data
-        for transceiver_entry in self.transceiver_entries:
+        for transceiver_entry, namespace in self.transceiver_entries:
             # extract interface name
             interface = transceiver_entry.split(mibs.TABLE_NAME_SEPARATOR_VBAR)[-1]
-            self._update_transceiver_cache(interface)
+            self._update_transceiver_cache(interface, namespace)
 
-    def _update_per_namespace_data(self, pubsub):
+    def _update_per_namespace_data(self, pubsub, db_instance):
```
qiluo-msft, line 184: >db_instance [](start = 49, length = 11)

db_index

src/sonic_ax_impl/mibs/ietf/rfc2737.py
SuvarnaMeenakshi: This function is currently in sonic_ax_impl/mibs/__init__.py

qiluo-msft: Sorry, I mean the section is only needed once after each time db connected.

```
        for i in range(len(self.statedb)):
            if not self.pubsub[i]:
                pattern = self.TRANSCEIVER_KEY_PATTERN
                self.pubsub[i] = mibs.get_redis_pubsub(self.statedb[i], self.statedb[i].STATE_DB, pattern)
```

---
In reply to: [451174442](https://github.com/Azure/sonic-snmpagent/pull/137#discussion_r451174442) [](ancestors = 451174442)

src/sonic_ax_impl/mibs/ietf/rfc2737.py
SuvarnaMeenakshi: updated as per comment.

SuvarnaMeenakshi:
Will raise a new PR to fix the comment:
I see a common pattern in all the functions like init_namespace_sync_d*. They just call the single db function and merge all returned items respectively.
How about abstract this high level function and return a dictionary of list, in this case the type is db -> { if_name_map, if_alias_map, if_id_map, oid_sai_map, oid_name_map, if_oid_namespace }
This comment is also true on dbs_keys_namespace

Time reduction and socket connection error is fixed in PR #140

PR #138, opened 2020-07-04
[Namespace]: Fix SAI_ID key used in cpfcIfTable and csqIfQosGroupStatsTable implementation

SuvarnaMeenakshi:
Signed-off-by: SuvarnaMeenakshi <sumeenak@microsoft.com>

**- What I did**
In multi-asic platform, SAI OID is not unique for the whole device. It is unique for an asic and within a single namespace.
This PR is to make sure that data structures that use SAI Object ID as a key to retrieve data from COUNTERS_DB updated so that a combination of SAI Object ID and port index is used as a key.
**- How I did it**
- Update data structure to use combination SAI Object ID and port index as a key to retrieve data from COUNTERS_DB.
- Update Unit-test mock DB in namespaces, to reflect the scenario where SAI Object is same across different namespaces.
- Updates done to MIB implementation  for the below MIB tables to get data from the right database instances:
  - cpfcIfTable
  - csqIfQosGroupStatsTable
  
**- How to verify it**
- 
**- Description for the changelog**

PR #139, opened 2020-07-09
[Namespace]: Fix to reduce time and fix counters related MIB

SuvarnaMeenakshi:
**- What I did**
- Fix add to reduce time in Namespace get_all, which is taking more time due to additional check to see if the key exists in a database.
- In multi-asic platform, sai id is unique only within namespace, so avoid using sai id as key.

**- How I did it**
- Remove key exists check in dbs_get_all
- Make changes to MIB implementation using SAI ID as key, to use interface index as key instead:
InterfacesMIB
InterfaceMIBObjects
cpfcIfTable
csqIfQosGroupStatsTable
- Make changes to rif/vlan related counters to use interface index as key instead of SAI ID.

**- How to verify it**


src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -41,9 +41,9 @@ def reinit_data(self):
         self.if_alias_map, \
         self.if_id_map, \
         self.oid_sai_map, \
-        self.oid_name_map = Namespace.init_namespace_sync_d_interface_tables(self.db_conn)
+        self.oid_name_map, _, _ = Namespace.init_namespace_sync_d_interface_tables(self.db_conn)
 
-        self.if_bpid_map = Namespace.dbs_get_bridge_port_map(self.db_conn, mibs.ASIC_DB)
+        self.if_bpid_map = Namespace.dbs_get_bridge_port_map(self.db_conn)
```
abdosi, line 46: why we don't need second Argument Here ?

src/sonic_ax_impl/mibs/ietf/rfc1213.py
```diff
@@ -196,34 +204,54 @@ def reinit_data(self):
...
+        for db_index in self.port_rif_map:
+            for rif_sai_id in self.rif_port_map[db_index]:
+                port_sai_id = self.rif_port_map[db_index][rif_sai_id]
+                if port_sai_id in self.lag_sai_oid_map[db_index]:
+                    if_idx = self.lag_sai_oid_map[db_index][port_sai_id]
+                elif port_sai_id in self.sai_oid_map[db_index]:
+                    if_idx = self.sai_oid_map[db_index][port_sai_id]
+                else:
+                    continue
+                rif_counters = self.db_conn[db_index].get_all(mibs.COUNTERS_DB, mibs.counter_table(rif_sai_id), blocking=True)
```
abdosi, line 243: why blocking is True in this case ?

tests/mock_tables/asic1/asic_db.json
```diff
@@ -1,11 +1,11 @@
 {
   "ASIC_STATE:SAI_OBJECT_TYPE_FDB_ENTRY:{\"bridge_id\":\"oid:0x0\",\"bridge_type\":\"SAI_FDB_ENTRY_BRIDGE_TYPE_1Q\",\"mac\":\"7C:FE:90:80:9F:10\",\"switch_id\":\"oid:0x21000000000000\",\"vlan\":\"1000\"}": {
-    "SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID": "oid:0x3a0000000006208",
+    "SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID": "oid:0x3a000000000616",
```
abdosi, line 3: what trigger all these below changes ?

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -373,7 +376,7 @@ def init_sync_d_queue_tables(db_conn):
 
     # Parse the queue_name_map and create the following maps:
     # port_queues_map -> {"if_index : queue_index" : sai_oid}
-    # queue_stat_map -> {queue stat table name : {counter name : value}}
+    # queue_stat_map -> {"if_index : queue stat table name" : {counter name : value}}
```
abdosi, line 379: Please add some comment as why these changes are needed (OID overlap) in Multi-Asic platforms

src/sonic_ax_impl/mibs/ietf/rfc1213.py
SuvarnaMeenakshi: The  previous code did a get_all with blocking=true, so kept the same. should not cause an issue because we get based on rif_sai_id from the specific db instance.

tests/mock_tables/asic1/asic_db.json
SuvarnaMeenakshi: The changes in mock tables are done so that the SAI ID for port and other objects are same in different namespace. Earlier, in the mock tables, it was unique across the device. To reflect the actual scenario where the same SAI Id can appear in a different namespace, the changes are made in asic_db and counter_db.

src/sonic_ax_impl/mibs/ietf/rfc1213.py
```diff
@@ -196,34 +204,54 @@ def reinit_data(self):
...
-            {sai_id: Namespace.dbs_get_all(self.db_conn, mibs.COUNTERS_DB, mibs.counter_table(sai_id), blocking=True)
-             for sai_id in rif_sai_ids}
+        self.if_counters = \
+            {if_idx: self.db_conn[self.if_oid_namespace[if_idx]].get_all(mibs.COUNTERS_DB, mibs.counter_table(self.oid_sai_map[if_idx]))
+            for if_idx in self.oid_sai_map}
+
+        for oid in self.vlan_oid_sai_map:
+            if oid in self.vlan_oid_namespace:
+                db_index = self.vlan_oid_namespace[oid]
+                self.rif_counters[oid] = self.db_conn[db_index].get_all(mibs.COUNTERS_DB, mibs.counter_table(self.vlan_oid_sai_map[oid]), blocking=True)
```
abdosi, line 232: the if condition will it not add to time taken ?

src/sonic_ax_impl/mibs/ietf/rfc1213.py
abdosi: but it will add to time taken ? same as original problem

src/sonic_ax_impl/mibs/__init__.py
SuvarnaMeenakshi: Updated as per comment.

src/sonic_ax_impl/mibs/ietf/rfc4363.py
SuvarnaMeenakshi: dbs_get_bridge_port_map() does not use the db name, the function only retrieves data from ASIC_DB, so did not have to use an argument to specify the db name.

qiluo-msft:
Remove extra file `foo`

SuvarnaMeenakshi:
Closing this PR as changes in this PR are not required.
Time increase and socket connection error is fixed using: https://github.com/Azure/sonic-snmpagent/pull/140
SAI ID fixes will be done in: https://github.com/Azure/sonic-snmpagent/pull/138

PR #140, opened 2020-07-10
[Namespace]: Fix SNMP AgentX  socket connection timeout when using Namespace.get_all() 

SuvarnaMeenakshi:

**- What I did**
To support namespaces, we added a Namespace class with functions to support getting data from all databases.
The Namespace dbs_get_all function , checks if the key exists in the database and does db connect before executing get_all.
Both of this operation causes additional time, which was causing agentx socket to disconnect.

**- How I did it**
- Remove the additional check that was added to see if the key exists before doing a get all.
- Remove db connect from dbs_get_all , as this was adding additional delay in dbs_get_all function, which is called during update_data. Instead, connect to all dbs during initialization, and connect to required dbs before invoking sync_d functions to ensure connection persists periodically.

**- How to verify it**
For single-asic platform and multi-asic, execute snmp queries to verify that the agentx socket does not disconnect and time issue is seen.

### DIFF
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -552,12 +552,12 @@ class Namespace:
         db get_all function executed on global and all namespace DBs.
         """
         result = {}
+        # If there are multiple namespaces, _hash might not be 
+        # present in all namespace, ignore if not present in a
+        # specfic namespace.
         if (len(dbs) > 1) : kwargs['blocking'] = False
         for db_conn in dbs:
             db_conn.connect(db_name)
-            # If there are multiple namespaces, _hash might not be 
-            # present in all namespace, ignore if not present in a
-            # specfic namespace.
             ns_result = db_conn.get_all(db_name, _hash, *args, **kwargs)
             if ns_result is not None:
                 result.update(ns_result)
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -552,7 +552,7 @@ class Namespace:
         # If there are multiple namespaces, _hash might not be 
         # present in all namespace, ignore if not present in a
         # specfic namespace.
-        if len(dbs) > 1: 
+        if len(dbs) > 1:
             kwargs['blocking'] = False
         for db_conn in dbs:
             ns_result = db_conn.get_all(db_name, _hash, *args, **kwargs)
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -553,9 +553,12 @@ class Namespace:
         # present in all namespace, ignore if not present in a
         # specfic namespace.
         if len(dbs) > 1:
-            kwargs['blocking'] = False
+            tmp_kwargs = kwargs.copy()
+            tmp_kwargs['blocking'] = False
+        else:
+            tmp_kwargs = kwargs
         for db_conn in dbs:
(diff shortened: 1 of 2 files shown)


src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -552,12 +552,16 @@ def dbs_get_all(dbs, db_name, _hash, *args, **kwargs):
...
         for db_conn in dbs:
-            db_conn.connect(db_name)
-            if(db_conn.exists(db_name, _hash)):
-                ns_result = db_conn.get_all(db_name, _hash, *args, **kwargs)
-                if ns_result is not None:
-                    result.update(ns_result)
+            # If there are multiple namespaces, _hash might not be 
+            # present in all namespace, ignore if not present in a
+            # specfic namespace.
+            ns_result = db_conn.get_all(db_name, _hash, blocking=False)
```
abdosi, line 562: just make it blocking = (db_conn.namespace == "")

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -552,12 +552,16 @@ def dbs_get_all(dbs, db_name, _hash, *args, **kwargs):
         db get_all function executed on global and all namespace DBs.
         """
         result = {}
+        Namespace.connect_all_dbs(dbs, db_name)
+        if len(dbs) == 1:
```
abdosi, line 556: No Need for this

src/sonic_ax_impl/mibs/__init__.py
abdosi, line 562: we should same parameters for get_all for single vs multi-asic call . 
we can update kwargs['blocking'] if present to False if len(dbs) > 1 (Muti-asic scenario)

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -552,12 +549,14 @@ def dbs_get_all(dbs, db_name, _hash, *args, **kwargs):
         db get_all function executed on global and all namespace DBs.
         """
         result = {}
+        # If there are multiple namespaces, _hash might not be 
+        # present in all namespace, ignore if not present in a
+        # specfic namespace.
+        if (len(dbs) > 1) : kwargs['blocking'] = False
```
qiluo-msft, line 555: >kwargs['blocking'] = False [](start = 28, length = 26)

Please it in next line.

qiluo-msft, line 555: >( [](start = 11, length = 1)

you can remove '()'

qiluo-msft, line 555: >kwargs [](start = 28, length = 6)

This parameter is passed as reference, and it is mutable.
If you change it inside function, it will have side-effect to the caller's scope. Could you do not change it?

abdosi: @qiluo-msft We were initially updating all the function call arguments and the changes were spread in multiple places. This approach looks better where only for Multi-Asic we are making blocking as False and for single asic behavior remains unchanged.
Also caller should not be impacted by this as it does not get used in caller scope and just passes the value to the library.

SuvarnaMeenakshi: fixed as per comment.

SuvarnaMeenakshi: Fixed as per comment.

SuvarnaMeenakshi: As per suggestion, instead of modifying kwargs directly, did a shallow copy to change.

PR #141, opened 2020-07-10
[Namespace] Fix interfaces counters in InterfacesMIB RFC 2863

SuvarnaMeenakshi:
**- What I did**
This PR is to fix InterfacesMIB rfc2863 implementation to make sure that interfaces  counters is keyed based on interface index and not SAI OID.

**- How I did it**
- Add namespace parameter in key ; instead of using SAI OID as key , the key will be namespace:sai_oid for multi-asic platform.
- In InterfacesMIB ; update code to use interface index as key in interface counter dict.
- Skip the unit-test cases for other MIBs where SAI OID is used, until the implementation is updated for multiple namespaces.
FDB/PFC and InterfacesMIB in rfc1213.

**- How to verify it**
Tested InterfacesMIB snmp query on multi-asic platform.

### DIFF
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -161,16 +161,21 @@ def mgmt_if_entry_table_state_db(if_name):
     return b'MGMT_PORT_TABLE|' + if_name
 
 def get_sai_id_key(namespace, sai_id):
-    if namespace != "":
-        return namespace.encode() + b':' + sai_id
+    sai_id = sai_id.decode()
+    if namespace != '':
+        key = namespace + ':' + sai_id
     else:
-        return sai_id
+        key = sai_id
+    return key.encode()
 
 def split_sai_id_key(sai_id_key):
-    if ':' in str(sai_id_key):
-        return sai_id_key.split(b':')[0].decode(), sai_id_key.split(b':')[1]
+    sai_id_key = sai_id_key.decode()
+    if ':' in sai_id_key:
+        namespace, sai_id = sai_id_key.split(':')
     else:
-        return "", sai_id_key
+        namespace = ''
+        sai_id = sai_id_key
+    return namespace, sai_id.encode()
 
 def config(**kwargs):
     global redis_kwargs
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -160,6 +160,17 @@ def mgmt_if_entry_table_state_db(if_name):
 
     return b'MGMT_PORT_TABLE|' + if_name
 
+def get_sai_id_key(namespace, sai_id):
+    if namespace != "":
+        return str(namespace) + ':' + sai_id
+    else:
+        return sai_id
+
+def split_sai_id_key(sai_id_key):
+    if ':' in sai_id_key:
+        return sai_id_key.split(':')[0], sai_id_key.split(':')[1]
+    else:
+        return "", sai_id_key
 
 def config(**kwargs):
     global redis_kwargs
@@ -221,7 +232,8 @@ def init_sync_d_interface_tables(db_conn):
     if_name_map = {if_name: sai_id for if_name, sai_id in if_name_map.items() if \
                    (re.match(port_util.SONIC_ETHERNET_RE_PATTERN, if_name.decode()) or \
                     re.match(port_util.SONIC_ETHERNET_BP_RE_PATTERN, if_name.decode()))}
-    if_id_map = {sai_id: if_name for sai_id, if_name in if_id_map.items() if \
+    # sai_id_key = namespace : sai_id
+    if_id_map = {get_sai_id_key(db_conn.namespace, sai_id): if_name for sai_id, if_name in if_id_map.items() if \
                  (re.match(port_util.SONIC_ETHERNET_RE_PATTERN, if_name.decode()) or \
                   re.match(port_util.SONIC_ETHERNET_BP_RE_PATTERN, if_name.decode()))}
     logger.debug("Port name map:\n" + pprint.pformat(if_name_map, indent=2))
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -564,12 +564,16 @@ class Namespace:
         db get_all function executed on global and all namespace DBs.
         """
         result = {}
+        Namespace.connect_all_dbs(dbs, db_name)
+        if len(dbs) == 1:
+            return dbs[0].get_all(db_name, _hash, *args, **kwargs)
         for db_conn in dbs:
-            db_conn.connect(db_name)
-            if(db_conn.exists(db_name, _hash)):
-                ns_result = db_conn.get_all(db_name, _hash, *args, **kwargs)
-                if ns_result is not None:
-                    result.update(ns_result)
+            # If there are multiple namespaces, _hash might not be 
+            # present in all namespace, ignore if not present in a
+            # specfic namespace.
+            ns_result = db_conn.get_all(db_name, _hash, blocking=False)
+            if ns_result is not None:
+                result.update(ns_result)
         return result
 
     @staticmethod
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -162,13 +162,13 @@ def mgmt_if_entry_table_state_db(if_name):
 
 def get_sai_id_key(namespace, sai_id):
     if namespace != "":
-        return str(namespace) + ':' + sai_id
+        return namespace.encode() + b':' + sai_id
     else:
         return sai_id
 
 def split_sai_id_key(sai_id_key):
-    if ':' in sai_id_key:
-        return sai_id_key.split(':')[0], sai_id_key.split(':')[1]
+    if ':' in str(sai_id_key):
+        return sai_id_key.split(b':')[0].decode(), sai_id_key.split(b':')[1]
     else:
         return "", sai_id_key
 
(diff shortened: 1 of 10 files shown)


src/sonic_ax_impl/mibs/ietf/rfc2863.py
```diff
@@ -4,6 +4,7 @@
 from sonic_ax_impl import mibs
 from ax_interface.mib import MIBMeta, MIBUpdater, ValueType, SubtreeMIBEntry, OverlayAdpaterMIBEntry, OidMIBEntry
 from sonic_ax_impl.mibs import Namespace
+from swsssdk.port_util import  get_index_from_str
```
qiluo-msft, line 7: >  [](start = 30, length = 1)

One extra blank

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -160,6 +160,22 @@ def mgmt_if_entry_table_state_db(if_name):
 
     return b'MGMT_PORT_TABLE|' + if_name
 
+def get_sai_id_key(namespace, sai_id):
+    sai_id = sai_id.decode()
```
qiluo-msft, line 164: >decode [](start = 20, length = 6)

decode and encode will applied to single ASIC and this could be optimized.

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -160,6 +160,22 @@ def mgmt_if_entry_table_state_db(if_name):
...
+    if namespace != '':
+        key = namespace + ':' + sai_id
```
qiluo-msft, line 166: >namespace + ':' + sai_id [](start = 14, length = 24)

why not just concat b':' ?

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -160,6 +160,22 @@ def mgmt_if_entry_table_state_db(if_name):
...
+    else:
+        key = sai_id
+    return key.encode()
+
+def split_sai_id_key(sai_id_key):
+    sai_id_key = sai_id_key.decode()
+    if ':' in sai_id_key:
+        namespace, sai_id = sai_id_key.split(':')
```
qiluo-msft, line 174: >split(':') [](start = 39, length = 10)

Why not just `split(b':')` without decoding?

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -221,7 +237,8 @@ def init_sync_d_interface_tables(db_conn):
     if_name_map = {if_name: sai_id for if_name, sai_id in if_name_map.items() if \
                    (re.match(port_util.SONIC_ETHERNET_RE_PATTERN, if_name.decode()) or \
                     re.match(port_util.SONIC_ETHERNET_BP_RE_PATTERN, if_name.decode()))}
-    if_id_map = {sai_id: if_name for sai_id, if_name in if_id_map.items() if \
+    # sai_id_key = namespace : sai_id
```
qiluo-msft, line 240: >: [](start = 29, length = 1)

This is a new assumption that ':' will not appear in name space or sai_id. Please add code comment.

qiluo-msft: How about using '|' which is widely used in Redis keys

---
In reply to: [453139692](https://github.com/Azure/sonic-snmpagent/pull/141#discussion_r453139692) [](ancestors = 453139692)

src/sonic_ax_impl/mibs/ietf/rfc2863.py
```diff
@@ -98,10 +99,12 @@ def update_data(self):
...
-        self.if_counters = {
-            sai_id: Namespace.dbs_get_all(self.db_conn, mibs.COUNTERS_DB, mibs.counter_table(sai_id), blocking=True)
-            for sai_id in self.if_id_map}
-
+        for sai_id_key in self.if_id_map:
+            namespace, sai_id = mibs.split_sai_id_key(sai_id_key)
+            if_idx = get_index_from_str(self.if_id_map[sai_id_key].decode())
+            for db in self.db_conn:
+                if db.namespace == namespace:
+                    self.if_counters[if_idx] = db.get_all(mibs.COUNTERS_DB, mibs.counter_table(sai_id), blocking=True)
```
qiluo-msft, line 107: >self.if_counters[if_idx] = db.get_all(mibs.COUNTERS_DB, mibs.counter_table(sai_id), blocking=True) [](start = 20, length = 98)

break after this line?

qiluo-msft:
>         db_conn= []

Add one blank before `=`

---
Refers to: src/sonic_ax_impl/mibs/__init__.py:540 in 14353e2. [](commit_id = 14353e2e13a65ee9ad8398788463cf5650f2140c, deletion_comment = False)

src/sonic_ax_impl/mibs/ietf/rfc2863.py
```diff
@@ -98,10 +99,12 @@ def update_data(self):
...
         """
-        self.if_counters = {
-            sai_id: Namespace.dbs_get_all(self.db_conn, mibs.COUNTERS_DB, mibs.counter_table(sai_id), blocking=True)
-            for sai_id in self.if_id_map}
-
+        for sai_id_key in self.if_id_map:
+            namespace, sai_id = mibs.split_sai_id_key(sai_id_key)
+            if_idx = get_index_from_str(self.if_id_map[sai_id_key].decode())
+            for db in self.db_conn:
+                if db.namespace == namespace:
```
qiluo-msft, line 106: This is not efficient, because you search `self.db_conn` too many times.

Propose 2 alternative solutions:
1. create a map (namespace -> db)
2. use db_index instread of namespace in `get_sai_id_key`

tests/namespace/test_fdb.py
```diff
@@ -23,6 +23,7 @@
 class TestSonicMIB(TestCase):
     @classmethod
     def setUpClass(cls):
+        cls.skipTest(cls, "Namespace not implemented")
```
qiluo-msft, line 26: >skipTest [](start = 12, length = 8)

What is the meaning of "Namespace not implemented"?
Why you skip test?

SuvarnaMeenakshi: For multi-asic platform, the change is made to if_id_map key to store namespace and sai id together.
with this change, MIBS using this key have to be updated to use the new type of key.
Instead of making this change on all MIBS together in same PR, will make change in other PRs.
So, for multi-asic platform, the unit-tests for these MIB implementations are skipped.

src/sonic_ax_impl/mibs/ietf/rfc2863.py
SuvarnaMeenakshi: Fixed

src/sonic_ax_impl/mibs/__init__.py
SuvarnaMeenakshi: Fixed.

src/sonic_ax_impl/mibs/__init__.py
SuvarnaMeenakshi: Modified as per comment.

src/sonic_ax_impl/mibs/__init__.py
SuvarnaMeenakshi: Modified as per comment.

src/sonic_ax_impl/mibs/__init__.py
SuvarnaMeenakshi: Added comment.

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -160,6 +160,31 @@ def mgmt_if_entry_table_state_db(if_name):
...
+    """
+    inputs:
+    namespace - string
+    sai id - bytes
+    Return type:
+    bytes
+    Return value: namespace:sai id or sai id
+    """
+    if namespace != '':
+        sai_id_key =  namespace + ':' + sai_id.decode()
```
qiluo-msft, line 173: >namespace + ':' + sai_id.decode() [](start = 22, length = 33)

`return namespace.encode() + b':' + sai_id`
This will be easier and shorter.

src/sonic_ax_impl/mibs/ietf/rfc2863.py
SuvarnaMeenakshi: Added a map, db_index cannot be obtained directly in this function and will have to  be passed as an additional argument.

src/sonic_ax_impl/mibs/__init__.py

src/sonic_ax_impl/mibs/ietf/rfc2863.py
abdosi: @qiluo-msft thanks for pointing out. 
we discussed to have map but to avoid new map went with this approach of checking namespace and break considering loop is small.

tests/namespace/test_fdb.py
abdosi: @SuvarnaMeenakshi better we should add similar comment in test case itself  as TODO so that is easy to understand and track

SuvarnaMeenakshi: Updated comment with TODO

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -160,6 +160,30 @@ def mgmt_if_entry_table_state_db(if_name):
...
+        return sai_id
+
+def split_sai_id_key(sai_id_key):
+    """
+    Input - bytes
+    Return namespace string and sai id in byte string.
+    """
+    if b':' in sai_id_key:
+        namespace, sai_id = sai_id_key.split(b':')
+        return namespace.decode(), sai_id
```
qiluo-msft, line 184: you can split first and check result len().
This will scan the  string only once.

SuvarnaMeenakshi: Updated as per comment.

qiluo-msft:
@SuvarnaMeenakshi Please fix 2 new alerts

PR #144, opened 2020-07-13
[201911] Fix interface counters in RFC1213

SuvarnaMeenakshi:

**- What I did**
This PR is to fix InterfacesMIB rfc1213 implementation to make sure that interfaces counters is keyed based on interface index and not SAI OID.

**- How I did it**
In rfc1213 ; update code to use interface index as key in interface counter dict.
Run unit-test for namespace as well.

**- How to verify it**
Verified on multi-asic and single asic platform, to check interface counters are correct.


abdosi:
retest this please

abdosi:
retest snmpagent please

SuvarnaMeenakshi:
retest this please.

PR #145, opened 2020-07-13
[Namespace]: Fix interface counters in RFC 1213

SuvarnaMeenakshi:

**- What I did**
This PR is to fix InterfacesMIB rfc1213 implementation to make sure that interfaces counters is keyed based on interface index and not SAI OID.

**- How I did it**
- Fix library functions to use namespace:sai-oid as key, instead of using sai-oid.
- Fix RFC 1213 to use the new key format.

**- How to verify it**
Test InterfacesMIB on single and multi-asic platform.
single-asic platform should see no change in output.
multi-asic platform should show right interface counter values.


qiluo-msft:
>         return  namespace.encode() + b':' + sai_id

One extra blank

---
Refers to: src/sonic_ax_impl/mibs/__init__.py:173 in 8ed0978. [](commit_id = 8ed097813d9c2d2124c12b5122c0d8aeadcf38b0, deletion_comment = False)

SuvarnaMeenakshi:
> > ```
> >     return  namespace.encode() + b':' + sai_id
> > ```
> 
> One extra blank
> 
> Refers to: src/sonic_ax_impl/mibs/**init**.py:173 in [8ed0978](https://github.com/Azure/sonic-snmpagent/commit/8ed097813d9c2d2124c12b5122c0d8aeadcf38b0). [](commit_id = [8ed0978](https://github.com/Azure/sonic-snmpagent/commit/8ed097813d9c2d2124c12b5122c0d8aeadcf38b0), deletion_comment = False)

Removed extra space.

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -355,7 +367,7 @@ def init_sync_d_lag_tables(db_conn):
         return lag_name_if_name_map, if_name_lag_name_map, oid_lag_name_map, lag_sai_map
 
     lag_sai_map = db_conn.get_all(COUNTERS_DB, b"COUNTERS_LAG_NAME_MAP")
-    lag_sai_map = {name: sai_id.lstrip(b"oid:0x") for name, sai_id in lag_sai_map.items()}
+    lag_sai_map = {name: get_sai_id_key(db_conn.namespace, sai_id.lstrip(b"oid:0x")) for name, sai_id in lag_sai_map.items()}
```
qiluo-msft, line 370: >lag_sai_map [](start = 4, length = 11)

This line is too long to read. Could you rewrite by a loop?

src/sonic_ax_impl/mibs/ietf/rfc1213.py
```diff
@@ -206,17 +208,19 @@ def update_data(self):
         Update redis (caches config)
         Pulls the table references for each interface.
         """
-        self.if_counters = \
-            {sai_id: Namespace.dbs_get_all(self.db_conn, mibs.COUNTERS_DB, mibs.counter_table(sai_id), blocking=True)
-            for sai_id in self.if_id_map}
+        for sai_id_key in self.if_id_map:
+            namespace, sai_id = mibs.split_sai_id_key(sai_id_key)
+            if_idx = get_index_from_str(self.if_id_map[sai_id_key].decode())
```
qiluo-msft, line 213: >decode [](start = 67, length = 6)

Why previous we don't need `decode()` ?

SuvarnaMeenakshi: previously we never used the if_name from if_id_map, now we use this to get the interface index.
if_id_map previously also had name in bytes.

src/sonic_ax_impl/mibs/__init__.py

src/sonic_ax_impl/mibs/ietf/rfc1213.py
qiluo-msft: If previously never used, can you `decode` at the beginning of `init_sync_d_interface_tables`? So multiple `decode` could be simplified.



---
In reply to: [454632465](https://github.com/Azure/sonic-snmpagent/pull/145#discussion_r454632465) [](ancestors = 454632465)

SuvarnaMeenakshi: Correction to what I mentioned before, it was not used in this file, but rfc4363 is using this.
mibs.get_index(self.if_id_map[port_id]) . mibs.get_index() - this takes a binary string and decodes internally.
Modifying it in the init(), will impact some other users of that library function.

qiluo-msft: Thanks for providing an exception. I checked that function, and it is also a great opportunity to simplify this way.

---
In reply to: [454665605](https://github.com/Azure/sonic-snmpagent/pull/145#discussion_r454665605) [](ancestors = 454665605,454632465)

SuvarnaMeenakshi: If we update in init_sync_d_interface_tables(), we will end up doing a decode during reinit_data(); here decode is done during update_data(), so performance will not change.

SuvarnaMeenakshi: Updated as per suggestion.

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -246,7 +246,7 @@ def init_sync_d_interface_tables(db_conn):
     # namespace to get a unique key. Assuming that ':' is not present in namespace
     # string or in sai id.
     # sai_id_key = namespace : sai_id
-    if_id_map = {get_sai_id_key(db_conn.namespace, sai_id): if_name for sai_id, if_name in if_id_map.items() if \
+    if_id_map = {get_sai_id_key(db_conn.namespace, sai_id): if_name.decode() for sai_id, if_name in if_id_map.items() if \
```
qiluo-msft, line 249: >if_name.decode() [](start = 60, length = 16)

Many duplicated calls inside this function.

SuvarnaMeenakshi: Have updated the code to reduce the decode().
I kept the if_name_map to use bytes if_name, as the user functions use it as bytes  string.
But I had to add new dictionaries to fix this.

PR #146, opened 2020-07-14
Update transceiver info DB key names

jleveque:
Update transceiver info DB key names to align with recent change here: https://github.com/Azure/sonic-platform-common/pull/97

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc2737.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc2737.py
@@ -37,10 +37,10 @@ class XcvrInfoDB(bytes, Enum):
     """
 
     TYPE              = b"type"
-    HARDWARE_REVISION = b"hardwarerev"
-    SERIAL_NUMBER     = b"serialnum"
-    MANUFACTURE_NAME  = b"manufacturename"
-    MODEL_NAME        = b"modelname"
+    HARDWARE_REVISION = b"hardware_rev"
+    SERIAL_NUMBER     = b"serial"
+    MANUFACTURE_NAME  = b"manufacturer"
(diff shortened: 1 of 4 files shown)


src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -37,10 +37,10 @@ class XcvrInfoDB(bytes, Enum):
...
 
     TYPE              = b"type"
-    HARDWARE_REVISION = b"hardwarerev"
-    SERIAL_NUMBER     = b"serialnum"
-    MANUFACTURE_NAME  = b"manufacturename"
-    MODEL_NAME        = b"modelname"
+    HARDWARE_REVISION = b"hardware_rev"
+    SERIAL_NUMBER     = b"serial"
+    MANUFACTURE_NAME  = b"manufacturer"
+    MODEL_NAME        = b"model"
```
qiluo-msft, line 43: It is better to move into a common place and shared with the repo sonic-platform-common.

jleveque: It would prevent the need for dependent changes like this in the future. Please feel free to open an issue.

qiluo-msft: Could you clarify "prevent the need for dependent changes like this in the future"? I think it will make it easier.

jleveque: This PR was necessary because the key names were updated in sonic-platform-common. If the two repos shared a common source for the data, this PR would not be necessary.

prsunny:
@abdosi

PR #153, opened 2020-07-23
[LLDP]: Modify OID index of LLDPRemTableUpdater MIB

SuvarnaMeenakshi:

**- What I did**
LLDPRemTableUpdater index in OID is a tuple of  (remote time mark, if index, remote device index).
```
For example: LLDP-MIB::lldpRemSysName.615.177.11 = STRING: remote_device_name
here , remote time mark is 615, interface index is 615, remote device index is 11.
```
 In case of multi-asic platform, when querying this MIB it can happen that same if index result is seen in SNMP walk, with a different remote time mark. 
```
For example: 
LLDP-MIB::lldpRemSysName.615.177.11 = STRING: remote_device_name
LLDP-MIB::lldpRemSysName.625.177.11 = STRING: remote_device_name
```
In the above walk, 177 interface index data appears twice.

**- How I did it**
To avoid showing same if index result in the  result, set remote time mark to 0, as this value is not significant, 

**- How to verify it**
Make sure the output of single asic platform shows the same result as before but with OID modified as: lldpRemSysName.0.if_index.remote_dev_id.
This will take effect in the entire LLDPRemTableUpdater table.

Make sure the output on mulit asic platform shows the result with unique interface index.

### DIFF
--- a/src/sonic_ax_impl/mibs/ieee802_1ab.py
+++ b/src/sonic_ax_impl/mibs/ieee802_1ab.py
@@ -430,7 +430,7 @@ class LLDPRemTableUpdater(MIBUpdater):
             if not lldp_kvs:
                 continue
             try:
-                time_mark = int(lldp_kvs[b'lldp_rem_time_mark'])
+                time_mark = 0
                 remote_index = int(lldp_kvs[b'lldp_rem_index'])
                 self.if_range.append((time_mark,
                                       if_oid,
--- a/src/sonic_ax_impl/mibs/ieee802_1ab.py
+++ b/src/sonic_ax_impl/mibs/ieee802_1ab.py
@@ -430,6 +430,11 @@ class LLDPRemTableUpdater(MIBUpdater):
             if not lldp_kvs:
                 continue
             try:
+                # OID index for this MIB consists of remote time mark, if_oid, remote_index.
+                # For multi-asic platform, it can happen that same interface index result 
+                # is seen in SNMP walk, with a different remote time mark.
+                # To avoid repeating the data of same interface index with different remote 
+                # time mark, remote time mark is made as 0 in the OID indexing.
                 time_mark = 0
                 remote_index = int(lldp_kvs[b'lldp_rem_index'])
                 self.if_range.append((time_mark,


mykolaf:
Hi @SuvarnaMeenakshi ,
I would like to better understand the cause of the issue we are trying to solve with this PR.
My guess is that we for some reason have duplicate entries for a single interface in APP DB's LLDP_ENTRY table?

SuvarnaMeenakshi:
> Hi @SuvarnaMeenakshi ,
> I would like to better understand the cause of the issue we are trying to solve with this PR.
> My guess is that we for some reason have duplicate entries for a single interface in APP DB's LLDP_ENTRY table?

Hi @mykolaf , It is not that there are duplicate entries for same single interface in LLDP_ENTRY table, but the additional data for an interface is only seen on SNMP walk output.
iso.0.8802.1.1.2.1.4.1.1.9
LLDP-MIB::lldpRemSysName.615.177.11 = STRING: remove_device_name
LLDP-MIB::lldpRemSysName.625.177.11 = STRING: remove_device_name

The OID last 3 digits are formed as: <remote time mark>. <Current device if_idx> . < remote device index>

This is my understanding of the issue. when we issue SNMP walk – 
1)	 this triggers continuous get_next() – get_next reads the data structure to form the next OID. 
2)	This data structure has the same tuple as the OID (<remote time mark>. <Current device if_idx> . < remote device index>)
3)	At this point we have (615,177,11)
4)	In between, update_data() timer kicks in, it updates the same data structure, with a different remote time mark.
5)	Get_next continues, and gets the next data of previous OID, which returns (625,177, 11) – same data but new time mark.

So data structure in SNMP code has the correct data, DB has the correct data, it is only SNMP walk that has the issue.
This issue is very evident in mulit-asic platform. But this issue might be seen on single-asic platform as well, if there is some delay.

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -430,7 +430,7 @@ def update_data(self):
             if not lldp_kvs:
                 continue
             try:
-                time_mark = int(lldp_kvs[b'lldp_rem_time_mark'])
+                time_mark = 0
```
abdosi, line 433: please add comment for the motivation behind this change so that in future we can go-back and understand

SuvarnaMeenakshi: Added comment.

qiluo-msft:
Retest this please

SuvarnaMeenakshi:
retest this please

SuvarnaMeenakshi:
retest license/cla please

SuvarnaMeenakshi:
The license/cla check continues to be in 'waiting for status'. Closing this PR, Will raise a new PR for the same diff to re-run license/cla check.

SuvarnaMeenakshi:
Raised PR: https://github.com/Azure/sonic-snmpagent/pull/155 for the same diff.

PR #154, opened 2020-07-28
[Namespace]: Simplify sync_d functions to use higher order

SuvarnaMeenakshi:
functions.

**- What I did**
Namespace class was added to support getting data from multiple namespaces.
This class includes some functions which get data from multiple namespaces and combine them.
This PR is to simplify the function of getting from separate functions and merge them.

**- How I did it**
Write a higher order function to invoke individual function to retrieve data from a single namespace.
The higher order function will retrieve the data and combine.

**- How to verify it**
No change in single and multi-asic snmp walk outputs.
Only implementation changes.
Make sure snmp walk outputs before change and after change are the same.

### DIFF
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -556,8 +556,8 @@ class Namespace:
         
     @staticmethod
     def get_sync_d_from_all_namespace(per_namespace_func, dbs):
-        # return list of dictionaries retrieved from per namespace functions
-        # map of index:dict
+        # return merged tuple of dictionaries retrieved from per
+        # namespace functions.
         result_map = {}
         # list of return values
         result_list = []
@@ -570,7 +570,7 @@ class Namespace:
                     result_map[idx].update(ns_tuple[idx])
         for idx, ns_tuple_dict in result_map.items():
             result_list.append(ns_tuple_dict)
-        return(result_list)
+        return result_list
 
     @staticmethod
     def dbs_get_bridge_port_map(dbs, db_name):
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -554,82 +554,23 @@ class Namespace:
         else:
             return dbs[1:]
         
-
     @staticmethod
-    def init_namespace_sync_d_interface_tables(dbs):
-        if_name_map = {}
-        if_alias_map = {}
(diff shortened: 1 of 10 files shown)


src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -554,82 +554,23 @@ def get_non_host_dbs(dbs):
...
-        return port_queues_map, queue_stat_map, port_queue_list_map
+            ns_tuple = per_namespace_func(db_conn)
+            for idx in range(len(ns_tuple)):
+                if idx not in result_map:
+                    result_map[idx] = ns_tuple[idx]
+                else:
+                    result_map[idx].update(ns_tuple[idx])
+        for idx, ns_tuple_dict in result_map.items():
+            result_list.append(ns_tuple_dict)
+        return(result_list)
```
qiluo-msft, line 573:
```suggestion
        return result_list
```

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -554,82 +554,23 @@ def get_non_host_dbs(dbs):
...
-
-        """
-        all_ns_db - will have db_conn to all namespace DBs and
-        global db. First db in the list is global db.
-        Ignore first global db to get queue tables if there
-        are multiple namespaces.
-        """
-        Namespace.connect_namespace_dbs(dbs)
+    def get_sync_d_from_all_namespace(per_namespace_func, dbs):
+        # return list of dictionaries retrieved from per namespace functions
```
qiluo-msft, line 559: >return list of dictionaries [](start = 10, length = 27)

You did not return list of dictionaries

SuvarnaMeenakshi: we return list of dictionaries, the dictionaries will be the same as return from individual sync_d functions, just that they are merged together.

SuvarnaMeenakshi: Probably, we can say "return tuple of dictionaries"

src/sonic_ax_impl/mibs/__init__.py
SuvarnaMeenakshi: Updated as per comment.

src/sonic_ax_impl/mibs/__init__.py
SuvarnaMeenakshi: Updated as per review comment.

PR #157, opened 2020-08-28
Get port configuration from ConfigDB instead of AppDB

qiluo-msft:
**- What I did**
Due to history issue, the configuration of port are moved from AppDB to ConfigDB, and the data in AppDB is just for backward-compatible. We need to migrate the snmpagent to access ConfigDB for port configuration such as admin_status/alias etc. However, the oper_status is still in AppDB.

Add test cases for InterfaceMIB.

**- How I did it**

**- How to verify it**

### DIFF
--- a/src/sonic_ax_impl/mibs/ieee802_1ab.py
+++ b/src/sonic_ax_impl/mibs/ieee802_1ab.py
@@ -185,15 +185,11 @@ class LocPortUpdater(MIBUpdater):
 
     def _get_if_entry(self, if_name):
         if_table = ""
-
-        # Once PORT_TABLE will be moved to CONFIG DB
-        # we will get entry from CONFIG_DB for all cases
-        db = mibs.APPL_DB
+        db = mibs.CONFIG_DB
         if if_name in self.if_name_map:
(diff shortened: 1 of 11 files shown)


src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -185,15 +185,11 @@ def reinit_data(self):
...
     def _get_if_entry(self, if_name):
         if_table = ""
-
-        # Once PORT_TABLE will be moved to CONFIG DB
-        # we will get entry from CONFIG_DB for all cases
-        db = mibs.APPL_DB
+        db = mibs.CONFIG_DB
         if if_name in self.if_name_map:
-            if_table = mibs.if_entry_table(if_name)
+            if_table = mibs.if_entry_table_config_db(if_name)
```
prsunny, line 190: Why we cannot get everything from APP_DB? As you mentioned, "`oper_status`" is only in APP_DB. Also if there is no config_db entry for `MTU`, then there is a default `9100 `MTU created only in APP_DB table.

PR #164, opened 2020-10-08
Fix: correctly handle that lldp_loc_man_addr contains only IPv6 address without IPv4 address

qiluo-msft:
Fixed https://github.com/Azure/sonic-snmpagent/issues/163

### DIFF
--- a/src/sonic_ax_impl/mibs/ieee802_1ab.py
+++ b/src/sonic_ax_impl/mibs/ieee802_1ab.py
@@ -337,17 +337,15 @@ class LLDPLocManAddrUpdater(MIBUpdater):
 
     def get_next(self, sub_id):
         right = bisect_right(self.man_addr_list, sub_id)
-        if right == len(self.man_addr_list):
+        if right >= len(self.man_addr_list):
             return None
         return self.man_addr_list[right]
 
-    def lookup(self, sub_id, callable):
-        if len(sub_id) == 0:
+    def man_addr_subtype(self, sub_id):
+        if sub_id in self.man_addr_list:
+            return ManAddrConst.man_addr_subtype_ipv4
+        else:
             return None
-        return callable(sub_id)
-
-    @staticmethod
-    def man_addr_subtype(sub_id): return ManAddrConst.man_addr_subtype_ipv4
 
     def man_addr(self, sub_id):
         """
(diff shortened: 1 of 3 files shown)


lguohan:
what is the miss here? can you add unit test to prevent future regression?

qiluo-msft:
The issue is that LLDP_LOC_CHASSIS.lldp_loc_man_addr has only one IPv6 address, which happens outside the snmp container. The fix is just to handle this unexpected data peacefully.

We are trying to figure out a way to unit test.

lguohan:
what changed? is this a regression?

qiluo-msft:
I have not investigate outside snmp container. I believe it is an image regression issue.

lguohan:
> The issue is that LLDP_LOC_CHASSIS.lldp_loc_man_addr has only one IPv6 address, which happens outside the snmp container

how many ipv6 addresses does this have before?

lguohan (approved):

qiluo-msft:
Sorry for misleading. The statement should be:
The issue is that LLDP_LOC_CHASSIS.lldp_loc_man_addr has no IPv4 address(es) at all but only one or more IPv6 address(es), which happens outside the snmp container. The fix is just to handle this unexpected data peacefully.

lguohan:
what is the situation before?

qiluo-msft:
In a good system, I see one IPv4 plus one IPv6 address.

lguohan:
make sense, can you put this into the issue and ask @gen-hwa to look at it further? why we are missing ipv4 address now?

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -334,12 +337,12 @@ def update_data(self):
 
     def get_next(self, sub_id):
         right = bisect_right(self.man_addr_list, sub_id)
-        if right == len(self.man_addr_list):
+        if right >= len(self.man_addr_list):
```
SuvarnaMeenakshi, line 340: Why is this change required?

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -317,6 +317,9 @@ def reinit_data(self):
                 if '.' in mgmt_ip:
                     mgmt_ip_sub_oid = (addr_subtype_sub_oid, *[int(i) for i in mgmt_ip.split('.')])
                     break
+            else:
```
SuvarnaMeenakshi, line 320: else block is not intended to match the right if block.

src/sonic_ax_impl/mibs/ieee802_1ab.py
qiluo-msft: This is not required. It just make code more robust.

---
In reply to: [503552127](https://github.com/Azure/sonic-snmpagent/pull/164#discussion_r503552127) [](ancestors = 503552127)

src/sonic_ax_impl/mibs/ieee802_1ab.py
qiluo-msft: This is for-else structure, which is a critical change in this PR. If there is no IPv4 address matched in the for loop, we need to return soon. Otherwise future statement will fail since mgmt_ip_sub_oid==None

---
In reply to: [503552324](https://github.com/Azure/sonic-snmpagent/pull/164#discussion_r503552324) [](ancestors = 503552324)

SuvarnaMeenakshi: got it, thank you.

SuvarnaMeenakshi (approved):

Issue #165, opened 2020-10-11
Add a test case for LLDP_LOC_CHASSIS.lldp_loc_man_addr has only one IPv6 address


qiluo-msft:
Add a test case for LLDP_LOC_CHASSIS.lldp_loc_man_addr has only one IPv6 address.
This is a follow up issue for https://github.com/Azure/sonic-snmpagent/pull/164

**Description**

**Steps to reproduce the issue:**
1.
2.
3.

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

    **Output of `show version`:**

    ```
    (paste your output here)
    ```

    **Attach debug file `sudo generate_dump`:**

    ```
    (paste your output here)
    ```

PR #167, opened 2020-10-15
Add a test case for LLDP_LOC_CHASSIS.lldp_loc_man_addr has only one IPv6 address

qiluo-msft:
Then fixed https://github.com/Azure/sonic-snmpagent/issues/165

### DIFF
--- a/tests/mock_tables/global_db/appl_db.json
+++ b/tests/mock_tables/global_db/appl_db.json
@@ -20,6 +20,6 @@
     "lldp_loc_sys_desc": "Gotta go Fast!",
     "lldp_loc_sys_cap_enabled": "28 00",
     "lldp_loc_sys_cap_supported": "28 00",
-    "lldp_loc_man_addr": "10.224.25.36,fe80::ce37:abff:feec:de9c"
+    "lldp_loc_man_addr": "fe80::ce37:abff:feec:de9c"
   }
 }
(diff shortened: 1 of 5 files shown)


tests/mock_tables/global_db/appl_db.json
```diff
@@ -20,6 +20,6 @@
     "lldp_loc_sys_desc": "Gotta go Fast!",
     "lldp_loc_sys_cap_enabled": "28 00",
     "lldp_loc_sys_cap_supported": "28 00",
-    "lldp_loc_man_addr": "10.224.25.36,fe80::ce37:abff:feec:de9c"
+    "lldp_loc_man_addr": "fe80::ce37:abff:feec:de9c"
```
SuvarnaMeenakshi, line 23: The same table is present in other namespace APP_DBs also, can you make the same change in those as well, though it is not used in this MIB. https://github.com/Azure/sonic-snmpagent/blob/master/tests/mock_tables/asic0/appl_db.json#L30

qiluo-msft: Resolved

---
In reply to: [505910755](https://github.com/Azure/sonic-snmpagent/pull/167#discussion_r505910755) [](ancestors = 505910755)

SuvarnaMeenakshi (approved):

PR #173, opened 2020-11-04
[rfc4363]Validate port_id for sonic-buildimage issue 5345

tim-rj:
Signed-off-by: tim-rj <sonic_rd@ruijie.com.cn>

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc4363.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc4363.py
@@ -70,6 +70,8 @@ class FdbUpdater(MIBUpdater):
             if bridge_port_id not in self.if_bpid_map:
                 continue
             port_id = self.if_bpid_map[bridge_port_id]
+            if port_id not in self.if_id_map:
+                continue
 
             vlanmac = self.fdb_vlanmac(fdb)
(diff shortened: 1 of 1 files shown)


pettershao-ragilenetworks:
help forward this, thanks!

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -70,6 +70,8 @@ def update_data(self):
             if bridge_port_id not in self.if_bpid_map:
                 continue
             port_id = self.if_bpid_map[bridge_port_id]
+            if port_id not in self.if_id_map:
```
qiluo-msft, line 73: >port_id not in self.if_id_map: [](http://example.com/codeflow?start=15&length=30)

What is the root cause of this issue?

qiluo-msft: Could you give more description about your topology and config? I guess your DUT has a PortChannel in Vlan, and there are traffic hit the port in PortChannel. If this is true, we recently fixed it by [https://github.com/Azure/sonic-snmpagent/pull/203](https://github.com/Azure/sonic-snmpagent/pull/203). Could you try latest master code?

PR #176, opened 2020-11-13
[RFC4292][Namespace]: Fix implementation of RouteUpdater for multi-asic platform

SuvarnaMeenakshi:
For multi-asic platforms, the default route was getting retrieved from only
the last namespace as the result dictionary was being written with the same key.
To avoid this, modify implementation to retrieve default routes from all
front end asic namespaces. Update implementation to ensure that only
external routes are displayed in the result.

**- What I did**
- Fix RouteUpdater implementation and add unit-test to test multiple namespaces.

**- How I did it**
- Fix RouteUpdater implementation to retrieve default routes from all front end asic namespaces.
- Fix implementation to ensure only external routes are displayed in the result.
- Add test case to test RFC4292 implementation for multiple namespaces scenario.

**- How to verify it**
- Unit-test cases should pass.
- Update snmp docker with the changes on single asic and multi-asic platform.
- On single asic platform SNMP walk of 1.3.6.1.2.1.4.24.4 should not change before and after the fix.
- On multi-asic platform, SNMP results should show external default route information.

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc4292.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc4292.py
@@ -5,6 +5,7 @@ from sonic_ax_impl.mibs import Namespace
 from ax_interface import MIBMeta, ValueType, MIBUpdater, SubtreeMIBEntry
 from ax_interface.util import ip2tuple_v4
 from bisect import bisect_right
+from sonic_py_common import multi_asic
 
 class RouteUpdater(MIBUpdater):
     def __init__(self):
@@ -49,26 +50,34 @@ class RouteUpdater(MIBUpdater):
             self.route_dest_list.append(sub_id)
             self.route_dest_map[sub_id] = self.loips[loip].packed
 
-        route_entries = Namespace.dbs_keys(self.db_conn, mibs.APPL_DB, "ROUTE_TABLE:*")
-        if not route_entries:
-            return
+        # Get list of front end asic namespaces for multi-asic platform.
+        # This list will be empty for single asic platform.
+        front_ns = multi_asic.get_all_namespaces()['front_ns']
 
-        for route_entry in route_entries:
-            routestr = route_entry.decode()
-            ipnstr = routestr[len("ROUTE_TABLE:"):]
-            if ipnstr == "0.0.0.0/0":
-                ipn = ipaddress.ip_network(ipnstr)
-                ent = Namespace.dbs_get_all(self.db_conn, mibs.APPL_DB, routestr, blocking=True)
-                nexthops = ent[b"nexthop"].decode()
-                ifnames = ent[b"ifname"].decode()
-                for nh, ifn in zip(nexthops.split(','), ifnames.split(',')):
-                    ## Ignore non front panel interfaces
-                    ## TODO: non front panel interfaces should not be in APPL_DB at very beginning
-                    ## This is to workaround the bug in current sonic-swss implementation
-                    if ifn == "eth0" or ifn == "lo" or ifn == "docker0": continue
-                    sub_id = ip2tuple_v4(ipn.network_address) + ip2tuple_v4(ipn.netmask) + (self.tos,) + ip2tuple_v4(nh)
-                    self.route_dest_list.append(sub_id)
-                    self.route_dest_map[sub_id] = ipn.network_address.packed
+        for db_conn in Namespace.get_non_host_dbs(self.db_conn):
+            # For multi-asic platform, proceed to get routes only for front end namespaces.
+            if db_conn.namespace != multi_asic.DEFAULT_NAMESPACE and \
+               db_conn.namespace not in front_ns:
+                continue
+            ipnstr = "0.0.0.0/0"
+            route_str = "ROUTE_TABLE:0.0.0.0/0"
+            route_entry  = db_conn.keys(mibs.APPL_DB, route_str)
+            if not route_entry:
+                continue
+            ipn = ipaddress.ip_network(ipnstr)
+            ent = db_conn.get_all(mibs.APPL_DB, route_str, blocking=True)
+            nexthops = ent[b"nexthop"].decode()
+            ifnames = ent[b"ifname"].decode()
+            for nh, ifn in zip(nexthops.split(','), ifnames.split(',')):
+                ## Ignore non front panel interfaces
+                ## TODO: non front panel interfaces should not be in APPL_DB at very beginning
+                ## This is to workaround the bug in current sonic-swss implementation
+                if ifn == "eth0" or ifn == "lo" or ifn == "docker0": continue
+                # Ignore internal asic routes
+                if multi_asic.is_port_channel_internal(ifn, db_conn.namespace): continue
+                sub_id = ip2tuple_v4(ipn.network_address) + ip2tuple_v4(ipn.netmask) + (self.tos,) + ip2tuple_v4(nh)
+                self.route_dest_list.append(sub_id)
+                self.route_dest_map[sub_id] = ipn.network_address.packed
 
         self.route_dest_list.sort()
 
--- a/src/sonic_ax_impl/mibs/ietf/rfc4292.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc4292.py
@@ -53,18 +53,17 @@ class RouteUpdater(MIBUpdater):
         # Get list of front end asic namespaces for multi-asic platform.
         # This list will be empty for single asic platform.
         front_ns = multi_asic.get_all_namespaces()['front_ns']
+        ipnstr = "0.0.0.0/0"
+        ipn = ipaddress.ip_network(ipnstr)
+        route_str = "ROUTE_TABLE:0.0.0.0/0"
 
         for db_conn in Namespace.get_non_host_dbs(self.db_conn):
             # For multi-asic platform, proceed to get routes only for front end namespaces.
-            if db_conn.namespace != multi_asic.DEFAULT_NAMESPACE and \
-               db_conn.namespace not in front_ns:
-                continue
-            ipnstr = "0.0.0.0/0"
-            route_str = "ROUTE_TABLE:0.0.0.0/0"
+            if front_ns and db_conn.namespace not in front_ns: continue
+            port_config = multi_asic.get_port_table(db_conn.namespace)
             route_entry  = db_conn.keys(mibs.APPL_DB, route_str)
             if not route_entry:
                 continue
-            ipn = ipaddress.ip_network(ipnstr)
             ent = db_conn.get_all(mibs.APPL_DB, route_str, blocking=True)
             nexthops = ent["nexthop"]
             ifnames = ent["ifname"]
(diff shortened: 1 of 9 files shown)


src/sonic_ax_impl/mibs/ietf/rfc4292.py
```diff
@@ -49,26 +50,34 @@ def update_data(self):
...
+        front_ns = multi_asic.get_all_namespaces()['front_ns']
+
+        for db_conn in Namespace.get_non_host_dbs(self.db_conn):
+            # For multi-asic platform, proceed to get routes only for front end namespaces.
+            if db_conn.namespace != multi_asic.DEFAULT_NAMESPACE and \
+               db_conn.namespace not in front_ns:
+                continue
+            ipnstr = "0.0.0.0/0"
+            route_str = "ROUTE_TABLE:0.0.0.0/0"
+            route_entry  = db_conn.keys(mibs.APPL_DB, route_str)
```
abdosi, line 64: do we need to check this ? db_conn.get_all should return empty in case entry does not exist in APP DB.
can we check.

src/sonic_ax_impl/mibs/ietf/rfc4292.py
```diff
@@ -49,26 +50,34 @@ def update_data(self):
...
+        # This list will be empty for single asic platform.
+        front_ns = multi_asic.get_all_namespaces()['front_ns']
+
+        for db_conn in Namespace.get_non_host_dbs(self.db_conn):
+            # For multi-asic platform, proceed to get routes only for front end namespaces.
+            if db_conn.namespace != multi_asic.DEFAULT_NAMESPACE and \
+               db_conn.namespace not in front_ns:
+                continue
+            ipnstr = "0.0.0.0/0"
+            route_str = "ROUTE_TABLE:0.0.0.0/0"
```
abdosi, line 63: suggest to move line 62, 63 and line 65 out of loop as one time initialization

src/sonic_ax_impl/mibs/ietf/rfc4292.py
```diff
@@ -49,26 +50,34 @@ def update_data(self):
...
+            ent = db_conn.get_all(mibs.APPL_DB, route_str, blocking=True)
+            nexthops = ent["nexthop"]
+            ifnames = ent["ifname"]
+            for nh, ifn in zip(nexthops.split(','), ifnames.split(',')):
+                ## Ignore non front panel interfaces
+                ## TODO: non front panel interfaces should not be in APPL_DB at very beginning
+                ## This is to workaround the bug in current sonic-swss implementation
+                if ifn == "eth0" or ifn == "lo" or ifn == "docker0": continue
+                # Ignore internal asic routes
+                if multi_asic.is_port_channel_internal(ifn, db_conn.namespace): continue
```
abdosi, line 77: check for backend ports also to be on safer side.

src/sonic_ax_impl/mibs/ietf/rfc4292.py
```diff
@@ -49,26 +50,34 @@ def update_data(self):
...
-                    self.route_dest_list.append(sub_id)
-                    self.route_dest_map[sub_id] = ipn.network_address.packed
+        # Get list of front end asic namespaces for multi-asic platform.
+        # This list will be empty for single asic platform.
+        front_ns = multi_asic.get_all_namespaces()['front_ns']
+
+        for db_conn in Namespace.get_non_host_dbs(self.db_conn):
+            # For multi-asic platform, proceed to get routes only for front end namespaces.
+            if db_conn.namespace != multi_asic.DEFAULT_NAMESPACE and \
+               db_conn.namespace not in front_ns:
```
abdosi, line 60: this looks easy to read:
if front_ns and db_conn.namespace not in front_ns:
    continue

what do you think ?

src/sonic_ax_impl/mibs/ietf/rfc4292.py
SuvarnaMeenakshi: db_conn.get_all will return empty if blocking is set to false.

src/sonic_ax_impl/mibs/ietf/rfc4292.py
SuvarnaMeenakshi: updated as per comment. could not make use of multi_asic.is_port_internal() directly, because if we pass a port channel name to this function, then it will throw an error : Unknown port name . So instead checking the port_config table directly.

src/sonic_ax_impl/mibs/ietf/rfc4292.py
SuvarnaMeenakshi: Updated as per comment

src/sonic_ax_impl/mibs/ietf/rfc4292.py
SuvarnaMeenakshi: yes, that looks easier, Updated as per comment.

SuvarnaMeenakshi:
retest this please.

SuvarnaMeenakshi:
retest this please

src/sonic_ax_impl/mibs/ietf/rfc4292.py
SuvarnaMeenakshi: Updated to directly execute get_all() with blocking set to false.

src/sonic_ax_impl/mibs/ietf/rfc4292.py
```diff
@@ -49,26 +50,36 @@ def update_data(self):
...
+        # Get list of front end asic namespaces for multi-asic platform.
+        # This list will be empty for single asic platform.
+        front_ns = multi_asic.get_all_namespaces()['front_ns']
+        ipnstr = "0.0.0.0/0"
+        ipn = ipaddress.ip_network(ipnstr)
+        route_str = "ROUTE_TABLE:0.0.0.0/0"
+
+        for db_conn in Namespace.get_non_host_dbs(self.db_conn):
+            # For multi-asic platform, proceed to get routes only for front end namespaces.
+            if front_ns and db_conn.namespace not in front_ns: continue
```
qiluo-msft, line 62: >front_ns [](start = 15, length = 8)

why you need to test `front_ns` in each loop? if it must be a list, you don't need to test.

qiluo-msft, line 62: >continue [](start = 63, length = 8)

Following the code convention nearby, move `continue` to next line.

src/sonic_ax_impl/mibs/ietf/rfc4292.py
```diff
@@ -49,26 +50,36 @@ def update_data(self):
...
+            port_config = multi_asic.get_port_table(db_conn.namespace)
```
qiluo-msft, line 63: >port_config [](start = 12, length = 11)

`port_table` ?

src/sonic_ax_impl/mibs/ietf/rfc4292.py
```diff
@@ -49,26 +50,36 @@ def update_data(self):
...
+            ifnames = ent["ifname"]
+            for nh, ifn in zip(nexthops.split(','), ifnames.split(',')):
+                ## Ignore non front panel interfaces
+                ## TODO: non front panel interfaces should not be in APPL_DB at very beginning
+                ## This is to workaround the bug in current sonic-swss implementation
+                if ifn == "eth0" or ifn == "lo" or ifn == "docker0": continue
+                # Ignore internal asic routes
+                if multi_asic.is_port_channel_internal(ifn, db_conn.namespace): continue
+                if ifn in port_config and \
+                   multi_asic.ROLE in port_config[ifn] and \
```
qiluo-msft, line 77: >                    [](start = 0, length = 19)

Suggest code format in [https://stackoverflow.com/a/5253378/2514803](https://stackoverflow.com/a/5253378/2514803)

tests/mock_tables/dbconnector.py
```diff
@@ -141,6 +141,52 @@ def keys(self, pattern='*'):
...
+        if 'PORT_TABLE' in db:
+            new_key = k[len('PORT_TABLE:'):]
+            port_table[new_key] = db[k]
+    return port_table
+        
+multi_asic.get_num_asics = mock_get_num_asics
+multi_asic.is_multi_asic = mock_is_multi_asic
+multi_asic.get_all_namespaces = mock_get_all_namespaces
+multi_asic.is_port_channel_internal = mock_is_port_channel_internal
+multi_asic.get_port_table = mock_get_port_table
```
qiluo-msft, line 189: Seems that you don't need modify this file, suggest create another file to mock multi_asic.

src/sonic_ax_impl/mibs/ietf/rfc4292.py
```diff
@@ -49,26 +50,35 @@ def update_data(self):
...
+        # This list will be empty for single asic platform.
+        front_ns = multi_asic.get_all_namespaces()['front_ns']
+        ipnstr = "0.0.0.0/0"
+        ipn = ipaddress.ip_network(ipnstr)
+        route_str = "ROUTE_TABLE:0.0.0.0/0"
+
+        for db_conn in Namespace.get_non_host_dbs(self.db_conn):
+            # For multi-asic platform, proceed to get routes only for front end namespaces.
+            if front_ns and db_conn.namespace not in front_ns: continue
+            port_config = multi_asic.get_port_table(db_conn.namespace)
```
qiluo-msft, line 63: >port_config [](start = 12, length = 11)

Could you filter the port_table with external ports only here?

qiluo-msft (changes requested):
As comments

src/sonic_ax_impl/mibs/ietf/rfc4292.py
qiluo-msft: OK, I understand. `front_ns` will be empty for single ASIC, so it will run the loop body.

---
In reply to: [523313340](https://github.com/Azure/sonic-snmpagent/pull/176#discussion_r523313340) [](ancestors = 523313340)

src/sonic_ax_impl/mibs/ietf/rfc4292.py
SuvarnaMeenakshi: There is another function "multi_asic.is_port_internal()" which will check for external port directly, because if we pass a port channel name to this function, then it will throw an error - "Unknown port name". So before invoking that function we will have to check if the interface name is of a port channel or port name.

qiluo-msft: I don't mean to change the function. Base on the `get_port_table()` result, you can filter based on role.

---
In reply to: [523329081](https://github.com/Azure/sonic-snmpagent/pull/176#discussion_r523329081) [](ancestors = 523329081)

SuvarnaMeenakshi: Right, but that will add a loop, to loop through all entries and get the role and get the sub-set.
Instead, we can just look for a specific port name once we have a port name and get the sub-role specific to that port.

src/sonic_ax_impl/mibs/ietf/rfc4292.py
SuvarnaMeenakshi: Added code comment to clarify about single ASIC.

SuvarnaMeenakshi: Updated as per comment.

src/sonic_ax_impl/mibs/ietf/rfc4292.py
SuvarnaMeenakshi: Updated as per comment.

src/sonic_ax_impl/mibs/ietf/rfc4292.py
SuvarnaMeenakshi: Updated as per comment.

tests/mock_tables/dbconnector.py
SuvarnaMeenakshi: Created a new file.

qiluo-msft (approved):
LGTM. Please also check other reviewers' comments.

abdosi (approved):
LGTM

xumia:
retest snmpagent please

abdosi:
retest this please

abdosi:
Included in 201911 via direct PR

PR #178, opened 2020-11-21
Set swsscommon logging level

qiluo-msft:
**Depends on https://github.com/Azure/sonic-swss-common/pull/418**
There are many warning messages in swsscommon. Here I set the log_level higher to ERROR, so warnings will disappear in syslog.

Fixed https://github.com/Azure/sonic-buildimage/issues/5992

### DIFF
--- a/src/sonic_ax_impl/__main__.py
+++ b/src/sonic_ax_impl/__main__.py
@@ -1,4 +1,5 @@
 from signal import signal, SIGUSR1
+import logging
 import logging.handlers
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


src/sonic_ax_impl/__main__.py
```diff
@@ -75,6 +76,10 @@ def install_fragments():
     sonic_ax_impl.logger.setLevel(log_level)
     ax_interface.logger.setLevel(log_level)
     swsssdk.logger.setLevel(log_level_sdk)
+    # Note: swsscommon uses a fixed log level for simplicity
+    # TODO: improve it when log_level option is used
+    log_level_swsscommon = swsscommon.Logger.SWSS_ERROR
```
SuvarnaMeenakshi, line 81: can this be modified as below:
if args.get('log_level') is not None and args.get('log_level') ==  logging.INFO:
    log_level_swsscommon  = swsscommon.Logger.SWSS_INFO
if args.get('log_level') is not None and args.get('log_level') ==  logging.ERROR:
    log_level_swsscommon  = swsscommon.Logger.SWSS_ERROR
else:
    log_level_swsscommon  = swsscommon.Logger.SWSS_ERROR

vaibhavhd: I think `if args.get('log_level') is not None` is not needed and we can directly check the value of `args.get('log_level')`


Qi, what are we achieving with this change (in order to fix Azure/sonic-buildimage#5992). Will the warnings start appearing as errors?

vaibhavhd: Can the description of the PR be modified with some context?

qiluo-msft: Warnings in swsscommon will still be warnings. Here I set the log_level higher to ERROR, so warning will disappear in syslog.

---
In reply to: [529841456](https://github.com/Azure/sonic-snmpagent/pull/178#discussion_r529841456) [](ancestors = 529841456)

qiluo-msft: Updated.

---
In reply to: [529842115](https://github.com/Azure/sonic-snmpagent/pull/178#discussion_r529842115) [](ancestors = 529842115)

SuvarnaMeenakshi (approved):

vaibhavhd (approved):
LGTM, thanks!

PR #186, opened 2020-12-22
[RouteUpdater]: Fix multi_asic mock function implementation and multi_asic variable name

SuvarnaMeenakshi:
[RouteUpdater]: Fix multi_asic mock function implementation.
Update multi_asic mock function to return port_table from config_db
was returning empty dictionary due to which RouteUpdater class
was not completely unit-tested as one of the condition check was
never set to true.

**- What I did**
1. Fix multi_asic mock function implementation to get port_table information.  There was a mistake in mock_get_port_table function due to which port_table was always empty dictionary.
2. Use multi_asic.get_port_table_for_asic() function as we require port_table only of a specific namespace and do not require the entire port_table.
3. Fix the incorrect multi_asic variable name used in RouteUdpater.
Provides fix for https://github.com/Azure/sonic-snmpagent/issues/188

**- How I did it**
1. Update multi_asic mock_get_port_table_for_asic implementation.
3. Fix the incorrect multi_asic variable name used in RouteUdpater.

**- How to verify it**
Load updated docker and query 1.3.6.1.2.1.4.24 MIB. Verify the result.
"AttributeError: module 'sonic_py_common.multi_asic' has no attribute 'ROLE'" error message should not be seen in syslog .

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc4292.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc4292.py
@@ -63,7 +63,7 @@ class RouteUpdater(MIBUpdater):
             # For single-asic platform, front_ns will be empty list.
             if front_ns and db_conn.namespace not in front_ns:
                 continue
-            port_table = multi_asic.get_port_table(db_conn.namespace)
+            port_table = multi_asic.get_port_table_for_asic(db_conn.namespace)
             ent = db_conn.get_all(mibs.APPL_DB, route_str, blocking=False)
             if ent is None:
                 continue
@@ -79,8 +79,8 @@ class RouteUpdater(MIBUpdater):
                 if multi_asic.is_port_channel_internal(ifn, db_conn.namespace):
                     continue
                 if (ifn in port_table and
-                   multi_asic.ROLE in port_table[ifn] and
-                   port_table[ifn][multi_asic.ROLE] == multi_asic.INTERNAL_PORT): 
+                   multi_asic.PORT_ROLE in port_table[ifn] and
+                   port_table[ifn][multi_asic.PORT_ROLE] == multi_asic.INTERNAL_PORT): 
                     continue
                 sub_id = ip2tuple_v4(ipn.network_address) + ip2tuple_v4(ipn.netmask) + (self.tos,) + ip2tuple_v4(nh)
                 self.route_dest_list.append(sub_id)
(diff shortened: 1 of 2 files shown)


src/sonic_ax_impl/mibs/ietf/rfc4292.py
```diff
@@ -79,8 +79,8 @@ def update_data(self):
                 if multi_asic.is_port_channel_internal(ifn, db_conn.namespace):
                     continue
                 if (ifn in port_table and
-                   multi_asic.ROLE in port_table[ifn] and
-                   port_table[ifn][multi_asic.ROLE] == multi_asic.INTERNAL_PORT): 
+                   multi_asic.PORT_ROLE in port_table[ifn] and
+                   port_table[ifn][multi_asic.PORT_ROLE] == multi_asic.INTERNAL_PORT):
```
qiluo-msft, line 83: >  [](start = 86, length = 1)

Remove trailing blank char.

qiluo-msft, line 83: If `if` statement is line breaked, you may add more indentation to them, so the branch code block is easier to find. There is a PEP.

SuvarnaMeenakshi: Updated as per comment.

SuvarnaMeenakshi: Updated as per comment.

qiluo-msft (approved):

abdosi:
@SuvarnaMeenakshi INTERNAL_PORT is already present in mock db ?

SuvarnaMeenakshi:
> @SuvarnaMeenakshi INTERNAL_PORT is already present in mock db ?

yes, for example:    https://github.com/Azure/sonic-snmpagent/blob/381ae478a6cb9f99e150406ed68cb672207bbddc/tests/mock_tables/asic0/config_db.json#L21
The mock_table's Role is not used to determine if a port channel is internal or not, to make the mock functions simpler, I have added a list of internal PortChannels here: https://github.com/Azure/sonic-snmpagent/blob/master/tests/mock_tables/multi_asic.py#L9
Also, there is a unit-test case to check if have an route through internal interface:
https://github.com/Azure/sonic-snmpagent/blob/381ae478a6cb9f99e150406ed68cb672207bbddc/tests/namespace/test_forward.py#L134

abdosi (approved):

PR #189, opened 2020-12-29
[rfc1213] fix counter value type.

stepanblyschak:
Counter values were stored as a mix of strings and ints.
This caused swsscommon to fail when there is an attempt
to write integer value where strings value was stored.

Signed-off-by: Stepan Blyshchak <stepanb@nvidia.com>

**- What I did**

**- How I did it**

**- How to verify it**


stepanblyschak:
@qiluo-msft please take a look

qiluo-msft (approved):

PR #193, opened 2021-01-26
Fix fdb_vlanmac() on corner cases

qiluo-msft:
**- What I did**
Fix fdb_vlanmac() on corner cases

**- How I did it**

**- How to verify it**
Existing unit test

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc4363.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc4363.py
@@ -81,6 +81,7 @@ class FdbUpdater(MIBUpdater):
             vlanmac = self.fdb_vlanmac(fdb)
             if not vlanmac:
                 mibs.logger.error("SyncD 'ASIC_DB' includes invalid FDB_ENTRY '{}': failed in fdb_vlanmac().".format(fdb_str))
+                continue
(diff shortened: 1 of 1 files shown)


src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -72,6 +79,8 @@ def update_data(self):
             port_id = self.if_bpid_map[bridge_port_id]
 
             vlanmac = self.fdb_vlanmac(fdb)
```
zhenggen-xu, line 81: should we just "try" these assignments, and capture exceptions instead of making branches in fdb_vlanmac().  And if fdb_vlanmac() is not successful, the rest assignment below should be ignored too.

qiluo-msft: Actually linters found the corner cases, and I fixed at the same time to make linting clean.
If you are worried about the CPU cycles, I tried minimize it for the normal cases.
I fixed the loop continue part.

---
In reply to: [564917045](https://github.com/Azure/sonic-snmpagent/pull/193#discussion_r564917045) [](ancestors = 564917045)

zhenggen-xu (approved):

SuvarnaMeenakshi (approved):

PR #196, opened 2021-02-09
[rfc2737] Fix issue: expect redis pubsub data to be str type instead of bytes type

Junchao-Mellanox:
**- What I did**

Change redis pubsub expect data type from bytes to str. There was a snmpagent framework change which transfer redis pubsub data from bytes to str, after that change, there is no need for non-framework code to do this transfer. So now we need handle str data directly.

**- How I did it**

Change data expectation from bytes to str

**- How to verify it**

Run regression and manual test

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc2737.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc2737.py
@@ -675,7 +675,7 @@ class PhysicalEntityCacheUpdater(object):
 
             db_entry = msg["channel"].split(":")[-1]
             data = msg['data']  # event data
-            if not isinstance(data, bytes):
+            if not isinstance(data, str):
                 continue
 
             # extract interface name


liat-grozovik:
@qiluo-msft can you please take a look?

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -675,7 +675,7 @@ def _update_per_namespace_data(self, pubsub):
 
             db_entry = msg["channel"].split(":")[-1]
             data = msg['data']  # event data
-            if not isinstance(data, bytes):
+            if not isinstance(data, str):
```
qiluo-msft, line 678: >str [](start = 36, length = 3)

This seems a valid fix. Wondering how it worked before? I think it hit a type runtime error below `"set" in data` ?

qiluo-msft (approved):

liat-grozovik:
@qiluo-msft  can we merge and set label to 202012

src/sonic_ax_impl/mibs/ietf/rfc2737.py
Junchao-Mellanox: it was b"set" before.

PR #200, opened 2021-03-11
[CI] Setup dummy azure pipeline

liushilongbuaa:
Signed-off-by: Shilong Liu <shilongliu@microsoft.com>

**- What I did**

**- How I did it**

**- How to verify it**


liushilongbuaa:
Former PR snmpagent does not have result. I think the reason is the same with tag missing issue of container sonic-slave-buster.

PR #203, opened 2021-03-13
Implement rfc4363 FdbUpdater for lag inside vlan

qiluo-msft:
**- What I did**
Currently if there are lags in vlan, and FDB entries with lag will not visible in FdbUpdater.

**- How I did it**
Implement rfc4363 FdbUpdater for lag inside vlan

**- How to verify it**
Test in DUT, and add unit test.


src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -304,13 +304,24 @@ def init_sync_d_lag_tables(db_conn):
...
     lag_entries = db_conn.keys(APPL_DB, "LAG_TABLE:*")
 
     if not lag_entries:
-        return lag_name_if_name_map, if_name_lag_name_map, oid_lag_name_map
+        return lag_name_if_name_map, if_name_lag_name_map, oid_lag_name_map, lag_sai_map
+
+    db_conn.connect(COUNTERS_DB)
+    lag_sai_map = db_conn.get_all(COUNTERS_DB, "COUNTERS_LAG_NAME_MAP")
+    for name, sai_id in lag_sai_map.items():
+        sai_id_key = get_sai_id_key(db_conn.namespace, sai_id.lstrip("oid:0x"))
```
SuvarnaMeenakshi, line 322: Can we use if_lag_name_map from port_util.if_lag_name_map() ?

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -31,7 +32,7 @@ def fdb_vlanmac(self, fdb):
                 self.bvid_vlan_map[fdb["bvid"]] = vlan_id
         else:
             return None
-        if not isinstance(vlan_id, str):
+        if not isinstance(vlan_id, bytes) and not isinstance(vlan_id, str):
```
SuvarnaMeenakshi, line 35: why is this change required?

SuvarnaMeenakshi (changes requested):
As comments.

src/sonic_ax_impl/mibs/__init__.py
qiluo-msft: The function `get_interface_oid_map` returns both port and lag. So not fit well.

---
In reply to: [594597081](https://github.com/Azure/sonic-snmpagent/pull/203#discussion_r594597081) [](ancestors = 594597081)

tests/test_fdb.py
```diff
@@ -56,6 +61,20 @@ def test_getnextpdu(self):
         self.assertEqual(value0.type_, ValueType.INTEGER)
         self.assertEqual(value0.data, 113)
 
+    def test_getnextpdu_lag(self):
```
abdosi, line 64: can we add same case for for multi-asic also . `tests/namespace/test_fdb.py`

SuvarnaMeenakshi: we currently skip the tests/namespace/test_fdb.py as fdb MIB is not modified to support for mulit-asic as it is a T0 topology feature.

abdosi: ok ideally we should remove test itself then.

abdosi (approved):
LGTM

src/sonic_ax_impl/mibs/ietf/rfc4363.py
qiluo-msft: I fixed it in above line, and this is not required anymore.

---
In reply to: [594597739](https://github.com/Azure/sonic-snmpagent/pull/203#discussion_r594597739) [](ancestors = 594597739)

SuvarnaMeenakshi (approved):

PR #204, opened 2021-03-15
Implement rfc4363 FdbUpdater for lag inside vlan

qiluo-msft:
**- What I did**
Same as https://github.com/Azure/sonic-snmpagent/pull/203
But target on 201911 branch.

**NOTICE**
Depends on https://github.com/Azure/sonic-py-swsssdk/pull/103

**- How I did it**

**- How to verify it**


qiluo-msft:
Retest this please

qiluo-msft:
Retest this please

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -77,12 +83,20 @@ def update_data(self):
             if bridge_port_id not in self.if_bpid_map:
                 continue
             port_id = self.if_bpid_map[bridge_port_id]
+            if port_id in self.if_id_map:
+                port_name = self.if_id_map[port_id]
+                port_index = mibs.get_index_from_str(port_name)
+            elif port_id in self.sai_lag_map:
+                port_name = self.sai_lag_map[port_id]
+                port_index = mibs.get_index(port_name)
```
abdosi, line 91: @qiluo-msft wondering show we keep `decode()` format of port_name in sai_lag_map same as what we do in for if_id_map https://github.com/Azure/sonic-snmpagent/blob/201911/src/sonic_ax_impl/mibs/__init__.py#L225. So that user of the map need not use different API `get_index_from_str()` and `get_index()`

abdosi:
@qiluo-msft PR description does not look correct. Need to be same as  master sonic-snmpagent PR.

qiluo-msft:
Updated PR description. Actually the titles are the same.

src/sonic_ax_impl/mibs/ietf/rfc4363.py
qiluo-msft: I checked all the items returned by `init_sync_d_lag_tables()`, they are all bytes. And this PR add one extra item there. It is weird to change the extra one as `str`.
If talking about master branch, there should be no such issue because all are `str`.

---
In reply to: [594477590](https://github.com/Azure/sonic-snmpagent/pull/204#discussion_r594477590) [](ancestors = 594477590)

abdosi (approved):

SuvarnaMeenakshi (approved):

PR #207, opened 2021-03-24
The ifIndex of based interface started in RFC1213 should be one

RayWang910012:
**- What I did**
remove "-1" in ifIndex return.
**- How I did it**

**- How to verify it**

The description of ifIndex in RFC1213 is

"A unique value for each interface.  
Its valueranges between 1 and the value of ifNumber.  
The value for each interface must remain constant at
least from one re-initialization of the entity's
network management system to the next reinitialization."

The based interface started from 0 is illegal in RFC1213 defined boundary.
So I think it should not minus one at this place.

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc1213.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc1213.py
@@ -266,7 +266,7 @@ class InterfacesUpdater(MIBUpdater):
         :return: the 0-based interface ID.
         """
         if sub_id:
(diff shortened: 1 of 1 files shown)


SuvarnaMeenakshi (approved):
LGTM

SuvarnaMeenakshi:
retest this please

qiluo-msft (changes requested):
LGTM. However, we need to discuss internally before merge.

PR #209, opened 2021-04-08
[multi-asic][RFC1213]: Oper status of mgmt interface for multi-asic platform

SuvarnaMeenakshi:
**- What I did**
Currently oper status of all management interface is retrieved from STATE DB.
Multi-asic platform does not have any daemon running on the host namespace to update oper status of management interface in STATE DB. This PR is to add a function to retrieve oper status of management interface from kernel file: "/sys/class/net/if_name/operstate".  
**- How I did it**
- Add a new function to retrieve oper status from /sys/class/net/if_name/operstate.
- Invoke this function periodically to retrieve oper status of all management interfaces.
- This is done periodically, instead of retrieving only on SNMP query, so that the time take during the SNMP query does not increase due to the file read operation.
- Add unit-tests for mgmt interface.
**- How to verify it**
- Run snmpwalk to get interface status, ensure that the interface status of management interface is correct, on single asic and multi-asic VS testbed.
- With the changes - no impact of single asic platform.
multi-asic VS testbed:
```
sudo ifconfig eth0 down
docker exec -it snmp snmpwalk -v2c -c public 127.0.0.1 1.3.6.1.2.1.2.2.1.8
iso.3.6.1.2.1.2.2.1.8.1 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.5 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.9 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.13 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.17 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.21 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.25 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.29 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.1001 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.1002 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.1003 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.1004 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.1005 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.1006 = INTEGER: 1
iso.3.6.1.2.1.2.2.1.8.5002 = INTEGER: 1
[...]

PR #210, opened 2021-04-08
[201911][multi-asic][RFC1213]: Oper status of mgmt interface for multi-asic platform

SuvarnaMeenakshi:
[201911][multi-asic][RFC1213]: Retrieve oper status of management
interface from kernel for multi-asic platform.
In multi-asic platform there is no daemon in the host
namespace to update the oper status of management interface.
Instead of retrieving this information from state_db,
retrieve this from /sys/class/net/../operstate file.

- What I did
This PR is same as : https://github.com/Azure/sonic-snmpagent/pull/209; for 201911
Currently oper status of all management interface is retrieved from STATE DB.
Multi-asic platform does not have any daemon running on the host namespace to update oper status of management interface in STATE DB. This PR is to add a function to retrieve oper status of management interface from kernel file: "/sys/class/net/if_name/operstate".
- How I did it

Add a new function to retrieve oper status from /sys/class/net/if_name/operstate.
Invoke this function periodically to retrieve oper status of all management interfaces.
This is done periodically, instead of retrieving only on SNMP query, so that the time take during the SNMP query does not increase due to the file read operation.
Add unit-tests for mgmt interface.
- How to verify it
Run snmpwalk to get interface status, ensure that the interface status of management interface is correct, on single asic and multi-asic VS testbed.
With the changes - no impact of single asic platform.


qiluo-msft:
Pending on https://github.com/Azure/sonic-snmpagent/pull/209

PR #169, opened 2020-10-16
[rfc1213] Interface MIB add l3 vlan interfaces & aggregate rif counters

stepanblyschak:
**- What I did**

Rework based on https://github.com/Azure/sonic-snmpagent/pull/133 and feedback https://github.com/Azure/sonic-snmpagent/issues/148.

**- How I did it**

- merge #133 with upstream master.
- Fix UnavailableDataError when RIF counters are not enabled.
- Make RIF keys in maps unique in multi asic env.

**- How to verify it**

- Disable rif counters
- restart swss
- snmpwalk
- Verify no UnavailableDataError in logs
- Verify port counters are returned
- Enable rif counters
- snmpwalk
- Verify no error in logs
- Verify error in/out counters are aggregated for rif ports and portchannels
- Verify VLAN RIF counters are present in the MIB.

**- Description for the changelog**

Added support for aggregated router interface counters and L3 VLAN counters to RFC1213 MIB.

PR #218, opened 2021-06-01
Vlan snmp

raphaelt-nvidia:
- What I did

Restored snmp vlan support per RFC1213, which had been reverted due to inconsistency with RFC2863, and added the missing support for RFC2863. Added unit tests for RFC2863.

- How I did it

Reverted https://github.com/Azure/sonic-snmpagent/pull/191, which was itself a revert of https://github.com/Azure/sonic-snmpagent/pull/169. Then added code to support vlan in rfc2863.py and unit tests.

Since a long time has elapsed since the original commit and other code has been pushed in the meantime, great care is needed in merging. Note in particular that mibs.init_sync_d_lag_tables now returns five parameters, the last two of which were added: lag_sai_map and sai_lap_map. This PR needs one of those maps, and a concurrent commit supporting RFC4363 needs the other, so all the callers were updated and use the one they need.

- How to verify it

Unit tests are run via make target/python-wheels/asyncsnmp-2.1.0-py3-none-any.whl.
See also Azure#169.

- Description for the changelog

Restored support for aggregated router interface counters and L3 VLAN counters to RFC1213 MIB, and extended to RFC2863.


dprital:
@lguohan - Who can review this PR ?

tests/test_interfaces.py
```diff
@@ -365,3 +365,580 @@ def test_in_octets_override(self):
...
+
+        response = get_pdu.make_response(self.lut)
+        print(response)
+
+        value0 = response.values[0]
+        self.assertEqual(value0.type_, ValueType.COUNTER_32)
+        self.assertEqual(str(value0.name), str(ObjectIdentifier(11, 0, 1, 0, (1, 3, 6, 1, 2, 1, 2, 2, 1, 20, 1001))))
+        self.assertEqual(value0.data, 106)
+
+class TestGetNextPDU_2863(TestCase):
```
qiluo-msft, line 858: >TestGetNextPDU_2863 [](http://example.com/codeflow?start=6&length=19)

Could you add another test case to make sure "RFC1213 is consistency with RFC2863"? This will also benefit future changes on either implementation or bug fix.

raphaelt-nvidia: Done

qiluo-msft (approved):

SuvarnaMeenakshi:
Can you modify changes added in this PR, so that interface description of Port Channel and management interface matches the value retrieved from config db, which will be empty string.
The below added code in "def interface_alias" in rfc2863.py; should be removed.

    if not result:
        #RFC2863 tables don't have descriptions for LAG, vlan & mgmt; take from RFC1213
        oid = self.get_oid(sub_id)
        if oid in self.oid_lag_name_map:
            result = self.oid_lag_name_map[oid]
        elif oid in self.mgmt_oid_name_map:
            result = self.mgmt_alias_map[self.mgmt_oid_name_map[oid]]
        elif oid in self.vlan_oid_name_map:
            result = self.vlan_oid_name_map[oid]
More information added here:
https://github.com/Azure/sonic-buildimage/pull/7859

PR #230, opened 2021-08-03
CPU Spike because of redundant and flooded keyspace notifis handled

vivekreddynv:
Signed-off-by: Vivek Reddy Karri <vkarri@nvidia.com>

**- What I did**

Fixes [#8293](https://github.com/Azure/sonic-buildimage/issues/8293)

**- How I did it**

Accumulated all the older notifications and did act only upon the latest notification discarding the others

**- How to verify it**

### DIFF
--- a/src/sonic_ax_impl/mibs/ieee802_1ab.py
+++ b/src/sonic_ax_impl/mibs/ieee802_1ab.py
@@ -96,6 +96,18 @@ def poll_lldp_entry_updates(pubsub):
         return ret
     return data, interface, if_index
 
+def get_latest_notification(pubsub):
+    """
+    Fetches the latest notification recorded on a lldp entry.
+    """
+    latest_update_map = {}
+    while True:
+        data, interface, if_index = poll_lldp_entry_updates(pubsub)
+        if not data:
+            break
+        latest_update_map[interface] = (data, if_index)
+    return latest_update_map
+        
 def parse_sys_capability(sys_cap):
     return bytearray([int (x, 16) for x in sys_cap.split()])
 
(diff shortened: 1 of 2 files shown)


vivekreddynv:
@qiluo-msft, please review

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -542,19 +554,17 @@ def _update_per_namespace_data(self, pubsub):
...
         Listen to updates in APP DB, update local cache
         """
-        while True:
-            data, interface, if_index = poll_lldp_entry_updates(pubsub)
-
-            if not data:
-                break
-
+        event_cache = get_latest_notification(pubsub)
+        for interface in event_cache.keys():
```
qiluo-msft, line 558: >.keys() [](http://example.com/codeflow?start=36&length=7)

You can remove `.keys()`

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -96,6 +96,18 @@ def poll_lldp_entry_updates(pubsub):
         return ret
     return data, interface, if_index
 
+def get_latest_notification(pubsub):
+    """
+    Fetches the latest notification recorded on a lldp entry.
+    """
+    latest_update_map = {}
+    while True:
+        data, interface, if_index = poll_lldp_entry_updates(pubsub)
```
qiluo-msft, line 105: >poll_lldp_entry_updates [](http://example.com/codeflow?start=36&length=23)

In future, pubsub may return just changed fields, not all the fields of a key. So the general solution should merge set/del entries in this function, instead of picking the last one.

vivekreddynv: Not sure if i understood your suggestion correctly, a keyspace notif doesn't contain any info about the fields changed. Eg: `__keyspace@0__:mykey del`. 

So, even if the lld_syncd logic was modified to send only one notif instead of mutiple. That'll be captured by this method. The actual logic of fetching the kv-pairs is in the downstream methods.

src/sonic_ax_impl/mibs/ieee802_1ab.py
vivekreddynv: Handled

src/sonic_ax_impl/mibs/ieee802_1ab.py
qiluo-msft: Thanks for clarify. I see the flow is implemented inside this file, instead of using a library. Sorry for misleading.

qiluo-msft (approved):

vivekreddynv:
@qiluo-msft, Can you backport this to 202106?

PR #232, opened 2021-08-28
[RFC1213]: Initialize lag oid map in reinit_data

SuvarnaMeenakshi:
**- What I did**
Initialize lag oid map in reinit_data instead of
updating in update_data. Updating lag oid map in update_data
can cause descrepancy as interface oid map is updated in
reinit_data.

There could be a scenario when lag oid map has lag oids updated in update_data but interfaces oid map is not updated as interfaces oid map is updated in reinit_data. When SNMP service comes up, there could be a short instance of time when interfaces oid map (oid_sai_map) is empty or not complete but lag oid map (oid_lag_name_map) is updated with lag oids and lag members. At this short span, if a SNMP query is done to get interfaces counters, current code will try to get LAG counters and will fail at https://github.com/Azure/sonic-snmpagent/blob/master/src/sonic_ax_impl/mibs/ietf/rfc1213.py#L383 when trying to get oid of lag members' oid. This can lead to key error : sai_id = self.oid_sai_map[oid]#012KeyError.
This change is to avoid this scenario.

**- How I did it**
init_sync_d_lag_tables in reinit_data instead of invoking in update_data.

**- How to verify it**
unit-test pass.
SNMP walk of interface MIB output is complete.

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc1213.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc1213.py
@@ -235,6 +235,11 @@ class InterfacesUpdater(MIBUpdater):
         self.rif_port_map, \
         self.port_rif_map = Namespace.get_sync_d_from_all_namespace(mibs.init_sync_d_rif_tables, self.db_conn)
 
+        self.lag_name_if_name_map, \
+        self.if_name_lag_name_map, \
+        self.oid_lag_name_map, \
+        self.lag_sai_map, self.sai_lag_map = Namespace.get_sync_d_from_all_namespace(mibs.init_sync_d_lag_tables, self.db_conn)
+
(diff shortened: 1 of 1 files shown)


qiluo-msft (approved):

qiluo-msft:
Could you also check which branches we need to backport?

SuvarnaMeenakshi:
> Could you also check which branches we need to backport?
This change is required in all previous branches 201811, 201911, 202106 and 202012
This change cannot be directly cherry picked to 201811 and 201911. So have created 2 separate PRs, 201911 change can be cherry picked to 202106 and 202012 branches

judyjoseph:
Not able to cherry-pick cleanly in 202106, please raise a separate PR for 202106.

PR #240, opened 2021-12-30
Modify path of python wheels to be installed.

SuvarnaMeenakshi:

**- What I did**
Modify path of python wheels to be installed as per changes in sonic-buildimage repo.
**- How I did it**
Modify path of swsssdk and sonic-py-common python wheels.

**- How to verify it**

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -46,8 +46,8 @@ stages:
         set -ex
         sudo apt-get -y purge libhiredis-dev libnl-3-dev libnl-route-3-dev
         sudo dpkg -i ../target/debs/buster/{libswsscommon_1.0.0_amd64.deb,python3-swsscommon_1.0.0_amd64.deb,libnl-3-200_*.deb,libnl-genl-3-200_*.deb,libnl-nf-3-200_*.deb,libnl-route-3-200_*.deb,libhiredis0.14_*.deb}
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -46,8 +46,8 @@ stages:
         set -ex
         sudo apt-get -y purge libhiredis-dev libnl-3-dev libnl-route-3-dev
         sudo dpkg -i ../target/debs/buster/{libswsscommon_1.0.0_amd64.deb,python3-swsscommon_1.0.0_amd64.deb,libnl-3-200_*.deb,libnl-genl-3-200_*.deb,libnl-nf-3-200_*.deb,libnl-route-3-200_*.deb,libhiredis0.14_*.deb}
-        sudo python3 -m pip install ../target/python-wheels/swsssdk*-py3-*.whl
-        sudo python3 -m pip install ../target/python-wheels/sonic_py_common-1.0-py3-none-any.whl
+        sudo python3 -m pip install ../target/python-wheels/buster/swsssdk*-py3-*.whl
```
qiluo-msft, line 49: >buster [](http://example.com/codeflow?start=60&length=6)

What is the reason that they are using different Debian version?
I think we should stick to the version same as whatever used in docker-snmp image.

SuvarnaMeenakshi: Fixed it, snmp is using buster debian version.

qiluo-msft (approved):

PR #245, opened 2022-03-07
Fix snmp agent Initialize config DB multiple times issue

liuh-80:
**- What I did**
    Fix following code issue:
        1. When initialize SonicDBConfig on multi ASIC device, not check if the global config already initialized issue.
        2. Initialize SonicDBConfig multiple times with dupe code.

**- How I did it**
    Code change to check isGlobalInit before load global config.
    Move dupe code to Namespace.init_sonic_db_config() and initialize config only once.
    Add new mock method for SonicDBConfig.isGlobalInit

**- How to verify it**
    Pass all UT and E2E test.

### DIFF
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -218,12 +218,14 @@ def init_db():
     Connects to DB
     :return: db_conn
     """
-    if not SonicDBConfig.isInit():
-        if multi_asic.is_multi_asic():
+    # SonicDBConfig has 2 initialize flags: isGlobalInit for multi asic and isInit for single asic
+    if multi_asic.is_multi_asic():
+        if not SonicDBConfig.isGlobalInit():
             # Load the global config file database_global.json once.
             SonicDBConfig.load_sonic_global_db_config()
-        else:
-            SonicDBConfig.load_sonic_db_config()
+    elif not SonicDBConfig.isInit():
+        SonicDBConfig.load_sonic_db_config()
+
     # SyncD database connector. THIS MUST BE INITIALIZED ON A PER-THREAD BASIS.
     # Redis PubSub objects (such as those within swsssdk) are NOT thread-safe.
     db_conn = SonicV2Connector(**redis_kwargs)
@@ -539,11 +541,12 @@ class Namespace:
     @staticmethod
     def init_namespace_dbs():
         db_conn = []
-        if not SonicDBConfig.isInit():
-            if multi_asic.is_multi_asic():
+        # SonicDBConfig has 2 initialize flags: isGlobalInit for multi asic and isInit for single asic
+        if multi_asic.is_multi_asic():
+            if not SonicDBConfig.isGlobalInit():
                 SonicDBConfig.load_sonic_global_db_config()
-            else:
-                SonicDBConfig.load_sonic_db_config()
+        elif not SonicDBConfig.isInit():
+            SonicDBConfig.load_sonic_db_config()
         host_namespace_idx = 0
         for idx, namespace in enumerate(SonicDBConfig.get_ns_list()): 
             if namespace == multi_asic.DEFAULT_NAMESPACE:
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -213,25 +213,6 @@ def config(**kwargs):
     redis_kwargs = {k:v for (k,v) in kwargs.items() if k in ['unix_socket_path', 'host', 'port']}
     redis_kwargs['decode_responses'] = True
 
-def init_db():
-    """
-    Connects to DB
-    :return: db_conn
-    """
-    # SonicDBConfig has 2 initialize flags: isGlobalInit for multi asic and isInit for single asic
-    if multi_asic.is_multi_asic():
-        if not SonicDBConfig.isGlobalInit():
-            # Load the global config file database_global.json once.
-            SonicDBConfig.load_sonic_global_db_config()
-    elif not SonicDBConfig.isInit():
-        SonicDBConfig.load_sonic_db_config()
-
-    # SyncD database connector. THIS MUST BE INITIALIZED ON A PER-THREAD BASIS.
-    # Redis PubSub objects (such as those within swsssdk) are NOT thread-safe.
-    db_conn = SonicV2Connector(**redis_kwargs)
-
-    return db_conn
-
 def init_mgmt_interface_tables(db_conn):
     """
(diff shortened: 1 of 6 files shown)


liuh-80:
> @liuh-80 Was there any error that you would see without this fix?

I'm working on a PR to change sonic_py_common from using swsssdk to swss_common, we found there are 2 issue with that change in this project:
1. UT crash, it's because the SonicDBConfig in swss_common has a new data member "isGlobalInit", but in the UT code, not mock this member, so I fix the UT issue in this PR.

2. snmp agent crash, there is a code bug in sonic_py_common device_info class, however during  check the crash issue we found in snmp agent there are dupe SonicDBConfig load config code in MIB code.

After discussion with Qi, we will improve the load config code, I will submit a new ituration later to move the dupe code to a common place and make it run one within SonicMIB class.

qiluo-msft:
> def init_db():

I think the key issue is to only initialize DBConfig once. Currently both init_namespace_dbs() and init_db() will try initialize.

---
Refers to: src/sonic_ax_impl/mibs/__init__.py:216 in 1070242. [](commit_id = 10702427e202deffb8829ed57c5bee7e0046bc60, deletion_comment = False)

liuh-80:
> > def init_db():
> 
> I think the key issue is to only initialize DBConfig once. Currently both init_namespace_dbs() and init_db() will try initialize.
> 
> Refers to: src/sonic_ax_impl/mibs/**init**.py:216 in [1070242](https://github.com/Azure/sonic-snmpagent/commit/10702427e202deffb8829ed57c5bee7e0046bc60). [](commit_id = [1070242](https://github.com/Azure/sonic-snmpagent/commit/10702427e202deffb8829ed57c5bee7e0046bc60), deletion_comment = False)

Fixed, move dupe code to Namespace.init_sonic_db_config() and initialize DB config in main() method.

liuh-80:
> > @liuh-80 Was there any error that you would see without this fix?
> 
> @SuvarnaMeenakshi, I'm working on a PR to change sonic_py_common from using swsssdk to swss_common, we found there are 2 issue with that change in this project:
> 
> 1. UT crash, it's because the SonicDBConfig in swss_common has a new data member "isGlobalInit", but in the UT code, not mock this member, so I fix the UT issue in this PR.
> 2. snmp agent crash, there is a code bug in sonic_py_common device_info class, however during  check the crash issue we found in snmp agent there are dupe SonicDBConfig load config code in MIB code.
> 
> After discussion with Qi, we will improve the load config code, I will submit a new ituration later to move the dupe code to a common place and make it run one within SonicMIB class.

@SuvarnaMeenakshi, I move the dupe code to Namespace.init_sonic_db_config() and initialize DB config in main() method. to  could you please review?

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -536,14 +519,35 @@ def get_oidvalue(self, oid):
...
 class Namespace:
+
+    @staticmethod
+    def init_sonic_db_config():
+        """
+        Initialize SonicDBConfig
+        """
+        # SonicDBConfig has 2 initialize flags: isGlobalInit for multi asic and isInit for single asic
+        if multi_asic.is_multi_asic():
+            if not SonicDBConfig.isGlobalInit():
```
qiluo-msft, line 530: >if not SonicDBConfig.isGlobalInit(): [](http://example.com/codeflow?start=12&length=36)

No need to check.

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -536,14 +519,35 @@ def get_oidvalue(self, oid):
...
+                # Load the global config file database_global.json once.
+                SonicDBConfig.load_sonic_global_db_config()
+        elif not SonicDBConfig.isInit():
```
qiluo-msft, line 533: >if not SonicDBConfig.isInit(): [](http://example.com/codeflow?start=10&length=30)

No need to check.

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -213,23 +213,6 @@ def config(**kwargs):
     redis_kwargs = {k:v for (k,v) in kwargs.items() if k in ['unix_socket_path', 'host', 'port']}
     redis_kwargs['decode_responses'] = True
 
-def init_db():
```
qiluo-msft, line 216: >init_db [](http://example.com/codeflow?start=4&length=7)

We could keep the function outside Namespace class, but internally call `Namespace.init_sonic_db_config()`

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -536,14 +519,35 @@ def get_oidvalue(self, oid):
         return self.oid_map[oid]
 
 class Namespace:
+
+    @staticmethod
+    def init_sonic_db_config():
```
qiluo-msft, line 524: >init_sonic_db_config [](http://example.com/codeflow?start=8&length=20)

We could add a static member `Namespace.db_config_loaded`.
1. init_sonic_db_config will check it and only load once
2. init_db will call init_sonic_db_config
3. init_namespace_dbs will call init_sonic_db_config
4. other caller do not need to call init_sonic_db_config explicitly

src/sonic_ax_impl/mibs/__init__.py
liuh-80: Fixed.

src/sonic_ax_impl/mibs/__init__.py
liuh-80: Fixed

src/sonic_ax_impl/mibs/__init__.py
liuh-80: Fixed, according to another comments, create a new method db_config_loaded()

src/sonic_ax_impl/mibs/__init__.py
liuh-80: Fixed, according to another comments, create a new method db_config_loaded()

src/sonic_ax_impl/mibs/__init__.py
qiluo-msft: Sorry for misleading, I mean a private member boolean variable Namespace.db_config_loaded. It will be more efficient.

liuh-80: Fixed, add a static class data member db_config_loaded

qiluo-msft (approved):

liuh-80:
@SuvarnaMeenakshi, could you please check this PR and give your comments?

SuvarnaMeenakshi (approved):

PR #250, opened 2022-04-01
[201811] Fix sonic_ax_impl memory leak in LLDPLocManAddrUpdater.

liuh-80:
**- What I did**
        Fix sonic_ax_impl memory leak in LLDPLocManAddrUpdater.

**- How I did it**
        In LLDPLocManAddrUpdater.reinit_data(), reset man_addr_list to empty list before append new data.

**- How to verify it**
        Pass all UT.
        Manually check there is no memory leak.

        Fix sonic_ax_impl memory leak in LLDPLocManAddrUpdater.

### DIFF
--- a/src/sonic_ax_impl/mibs/ieee802_1ab.py
+++ b/src/sonic_ax_impl/mibs/ieee802_1ab.py
@@ -295,11 +295,12 @@ class LLDPLocManAddrUpdater(MIBUpdater):
         """
         Subclass update data routine.
         """
+        self.man_addr_list = []
+
         # establish connection to application database.
         self.db_conn.connect(mibs.APPL_DB)
         mgmt_ip_bytes = self.db_conn.get(mibs.APPL_DB, mibs.LOC_CHASSIS_TABLE, b'lldp_loc_man_addr')
 
-        self.man_addr_list = []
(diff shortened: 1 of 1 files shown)


liuh-80:
Because sonic-snmpagent is a submodule of sonic-buildimage, so we need a submodule update PR in sonic-buildimage repo.

src/sonic_ax_impl/mibs/ieee802_1ab.py
```diff
@@ -299,6 +299,7 @@ def reinit_data(self):
         self.db_conn.connect(mibs.APPL_DB)
         mgmt_ip_bytes = self.db_conn.get(mibs.APPL_DB, mibs.LOC_CHASSIS_TABLE, b'lldp_loc_man_addr')
 
+        self.man_addr_list = []
```
qiluo-msft, line 302: Let's try to follow master branch as much as possible to prevent future conflict if we later decide cherry-pick something. Move this line to the beginning of this function, after the heredoc, and add one more empty line after.

liuh-80: Fixed.

qiluo-msft (approved):

PR #255, opened 2022-04-30
Fix: not to use blocking get_all() after keys()

qiluo-msft:
**- What I did**
In a dynamic environment, it is possible that some of the keys may
disappear between invoking keys() and get_all().

Prevent unnecessary timeout of blocking get_all().

**- How I did it**

**- How to verify it**

### DIFF
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -630,7 +630,7 @@ class Namespace:
         result = {}
         # If there are multiple namespaces, _hash might not be
         # present in all namespace, ignore if not present in a
-        # specfic namespace.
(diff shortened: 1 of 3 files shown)


SuvarnaMeenakshi:
Going ahead, how do we figure out if we have to use blocking call or not for a given Table in DB?
Are we planning to change all blocking calls to non-blocking?

SuvarnaMeenakshi:
> Going ahead, how do we figure out if we have to use blocking call or not for a given Table in DB? Are we planning to change all blocking calls to non-blocking?
Closing this comment, fixing only in functions which have keys/get_all calls.

SuvarnaMeenakshi (approved):
lgtm

PR #272, opened 2022-11-13
Remove error logging on "failed in fdb_vlanmac"

qiluo-msft:
**- What I did**
It is considered normal that the dependences of FDB_ENTRY are temporarily not available.

**- How I did it**

**- How to verify it**

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc4363.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc4363.py
@@ -98,9 +98,6 @@ class FdbUpdater(MIBUpdater):
                 continue
 
             vlanmac = self.fdb_vlanmac(fdb)
-            if not vlanmac:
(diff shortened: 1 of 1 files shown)


src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -99,7 +99,6 @@ def update_data(self):
 
             vlanmac = self.fdb_vlanmac(fdb)
             if not vlanmac:
-                mibs.logger.error("SyncD 'ASIC_DB' includes invalid FDB_ENTRY '{}': failed in fdb_vlanmac().".format(fdb_str))
```
zhenggen-xu, line 102: Can we keep this as logger.debug so we don't log by default but if we have verbose log turned on we can still see it? 
BTW:  Do we know in what conditions the FDB entry could be invalid?

prsunny (dismissed):
agree with @zhenggen-xu to have it at debug level.

zhenggen-xu (approved):

PR #275, opened 2022-12-09
[chassis]: Register only relevant MIBs on Supervisor

SuvarnaMeenakshi:
This is done to reduce CPU usage of snmp_ax_impl to avoid seeing socket connection failure seen during snmp query sent to supervisor of Chassis platform.

**- What I did**
With packet-chassis Supervisor, we are seeing an issue where sending a SNMP query causes snmp_ax_impl-snmpd socket connection to break.
This is causing the SNMP result to error out with packet error or timeout.

This is because with packet-chassis Supervisor, we have 16 asics and we have interfaces data coming from each asic.
With periodic polling of all these interfaces data from each namespace is causing snmp agent to be busy and processing query and getting periodic data is causing snmp socket connection to break with snmpd.

To avoid this, this PR is to register only the MIBs that are relevant on Supervisor for Chassis.
We only need interfaces/CPU/platform modules related information. We will not need the LLDP/Route/BGP related information from Supervisor or they will not even be present.
Though the data is not present, many MIB periodic function rely on getting all interfaces data (sync_d functions).
By only registering the relevant MIBs, I could see the CPU usage of snmp_ax_impl going down and socket connection also does not break.
**- How I did it**
If device is supervisor, register MIBs relevant for Supervisor.
**- How to verify it**
Verified that packet-chassis sup does not see any socket connection error.
No change on - single asic/multi-asic/linecard modules.

### DIFF
--- a/src/sonic_ax_impl/main.py
+++ b/src/sonic_ax_impl/main.py
@@ -13,6 +13,7 @@ from sonic_ax_impl.mibs import ieee802_1ab, Namespace
 from . import logger
 from .mibs.ietf import rfc1213, rfc2737, rfc2863, rfc3433, rfc4292, rfc4363
 from .mibs.vendor import dell, cisco
+from sonic_py_common import device_info
 
 # Background task update frequency ( in seconds )
 DEFAULT_UPDATE_FREQUENCY = 5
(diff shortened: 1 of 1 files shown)


SuvarnaMeenakshi:
Changes were made in syncd to fix overall CPU usage in packet-chassis Supervisor.
After that, we do not see the snmp socket connection breaking.
Closing this now.

PR #276, opened 2023-02-04
Fix: zero route may have empty nexthop

qiluo-msft:
**- What I did**
Fix: zero route may have empty nexthop. Improve code robustness.

**- How I did it**

**- How to verify it**
Manual test:
1. manipulate ApplDB default route: remove nexthop field and value
2. recover
3. manipulate ApplDB default route: remove ifname field and value
4. recover

Also add unit test.

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc1213.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc1213.py
@@ -153,13 +153,16 @@ class NextHopUpdater(MIBUpdater):
                 ipn = ipaddress.ip_network(ipnstr)
                 ent = Namespace.dbs_get_all(self.db_conn, mibs.APPL_DB, routestr, blocking=False)
                 if ent:
-                    nexthops = ent.get("nexthop", "")
-                    for nh in nexthops.split(','):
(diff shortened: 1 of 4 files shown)


src/sonic_ax_impl/mibs/ietf/rfc1213.py
```diff
@@ -153,7 +153,7 @@ def update_data(self):
                 ipn = ipaddress.ip_network(ipnstr)
                 ent = Namespace.dbs_get_all(self.db_conn, mibs.APPL_DB, routestr, blocking=False)
                 if ent:
-                    nexthops = ent["nexthop"]
+                    nexthops = ent.get("nexthop", "")
```
SuvarnaMeenakshi, line 156: If we return default empty string, the below for loop will get executed once.
```
>>> foo = ""
>>> for bar in foo.split(','):
...   print("hello")
... 
hello
```
Can we default to None and add additional check to execute the for loop only if not None.

SuvarnaMeenakshi (dismissed):
lgtm

SuvarnaMeenakshi (approved):

PR #281, opened 2023-05-21
SNMP hardening - redis socket connect

ycoheNvidia:
**- What I did**
in order to support snmp container hardening in https://github.com/sonic-net/sonic-buildimage/pull/15176
Made snmp agent to communicate with redis using socket
**- How I did it**
Added the use of socket connection to redis during snmp agent init

**- How to verify it**
Configure and run snmp
Added the use of socket connection to redis during snmp agent init

### DIFF
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -210,7 +210,9 @@ def split_sai_id_key(sai_id_key):
 
 def config(**kwargs):
     global redis_kwargs
-    redis_kwargs = {k:v for (k,v) in kwargs.items() if k in ['unix_socket_path', 'host', 'port', 'use_unix_socket_path']}
+    redis_kwargs = {k:v for (k,v) in kwargs.items() if k in ['unix_socket_path', 'host', 'port']}
+    if 'unix_socket_path' in redis_kwargs and 'host' in redis_kwargs and 'port' in redis_kwargs:
+        redis_kwargs['use_unix_socket_path'] = True
     redis_kwargs['decode_responses'] = True
 
 def init_db():
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -212,6 +212,7 @@ def config(**kwargs):
     global redis_kwargs
     redis_kwargs = {k:v for (k,v) in kwargs.items() if k in ['unix_socket_path', 'host', 'port']}
     redis_kwargs['decode_responses'] = True
+    redis_kwargs['use_unix_socket_path'] = True
 
 def init_db():
     """
--- a/src/sonic_ax_impl/mibs/__init__.py
+++ b/src/sonic_ax_impl/mibs/__init__.py
@@ -210,9 +210,8 @@ def split_sai_id_key(sai_id_key):
 
 def config(**kwargs):
     global redis_kwargs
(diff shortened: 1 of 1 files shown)


src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -212,6 +212,7 @@ def config(**kwargs):
     global redis_kwargs
     redis_kwargs = {k:v for (k,v) in kwargs.items() if k in ['unix_socket_path', 'host', 'port']}
     redis_kwargs['decode_responses'] = True
+    redis_kwargs['use_unix_socket_path'] = True
```
SuvarnaMeenakshi, line 215: Where is this function getting used?

https://github.com/sonic-net/sonic-snmpagent/blob/b2aca442672713093a15b9b4054df8cb15149d73/src/sonic_ax_impl/mibs/__init__.py#L571 already uses unix_socket_path.
db = SonicV2Connector(use_unix_socket_path=True, namespace=namespace)

qiluo-msft: I think we need to replace all `init_db` to `init_namespace_dbs`, and deprecate `init_db`.

ycoheNvidia: @SuvarnaMeenakshi I will check if line 215 can be added as part of list items in line 213, thanks

ycoheNvidia: done

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -210,7 +210,7 @@ def split_sai_id_key(sai_id_key):
 
 def config(**kwargs):
```
qiluo-msft, line 211: >config [](http://example.com/codeflow?start=4&length=6)

this process is general service, which support both TCP port and unix socket. and the choice is on the command line.

We can make preference of `unix_socket_path` above `host`/`port`, if all 3 are provided. No need to add `use_unix_socket_path` into service command line.

ycoheNvidia: Fixed - I hope it is ok

qiluo-msft (approved):

SuvarnaMeenakshi (approved):

qiluo-msft:
Could you add unit test to cover the code change?

src/sonic_ax_impl/mibs/__init__.py
```diff
@@ -211,6 +211,8 @@ def split_sai_id_key(sai_id_key):
...
     global redis_kwargs
     redis_kwargs = {k:v for (k,v) in kwargs.items() if k in ['unix_socket_path', 'host', 'port']}
+    if 'unix_socket_path' in redis_kwargs and 'host' in redis_kwargs and 'port' in redis_kwargs:
+        redis_kwargs['use_unix_socket_path'] = True
```
SuvarnaMeenakshi, line 215: what happens if unix_socket_path or host is not present in redis_kwargs?
I see that with master branch image, redis_kwargs only has 'decode_responses' and with that the connection to redis fails.
If I add "use_unix_socket_path" to true in redis_kwargs, then the connection is successful and snmp_subagent comes up. Can you re-check if this ?

ycoheNvidia:
> Could you add unit test to cover the code change?

We had a discussion about it few months ago, is there infra for it?

PR #286, opened 2023-07-21
Fix FdbUpdater crash when SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID attribute missing.

liuh-80:
Fix FdbUpdater crash when SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID attribute missing.

#### Work item tracking
Microsoft ADO (number only): 16189251

**- What I did**
Fix FdbUpdater crash when SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID attribute missing.

**- How I did it**
Add try-catch block in rfc3463 FdbUpdater

**- How to verify it**
Manually test.
Pass all UT

Fix FdbUpdater crash when SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID attribute missing.

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc4363.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc4363.py
@@ -61,6 +61,7 @@ class FdbUpdater(MIBUpdater):
 
         self.if_bpid_map = Namespace.dbs_get_bridge_port_map(self.db_conn, mibs.ASIC_DB)
         self.bvid_vlan_map.clear()
+        self.broken_fdbs.clear()
 
     def update_data(self):
         """
@@ -90,9 +91,9 @@ class FdbUpdater(MIBUpdater):
             try:
                 bridge_port_id_attr = ent["SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID"]
             except KeyError as e:
-                # Only write error log once
+                # Only write warning log once
                 if fdb_str not in self.broken_fdbs:
-                    mibs.logger.error("SyncD 'ASIC_DB' includes invalid FDB_ENTRY '{}': failed to get bridge_port_id, exception: {}".format(fdb_str, e))
+                    mibs.logger.warn("SyncD 'ASIC_DB' includes invalid FDB_ENTRY '{}': failed to get bridge_port_id, exception: {}".format(fdb_str, e))
                     self.broken_fdbs.append(fdb_str)
                 continue
 
--- a/src/sonic_ax_impl/mibs/ietf/rfc4363.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc4363.py
@@ -20,6 +20,7 @@ class FdbUpdater(MIBUpdater):
         self.vlanmac_ifindex_list = []
         self.if_bpid_map = {}
         self.bvid_vlan_map = {}
+        self.broken_fdbs = []
 
     def fdb_vlanmac(self, fdb):
         if 'vlan' in fdb:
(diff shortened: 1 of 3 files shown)


liuh-80:
Agent no disk space, close re-open to run validation again.

liuh-80:
PR validation failed because pipeline trying to download 17GB data to agent, and most agent does not have enough space:

Minimatch patterns: [**]
Filtered 1867 files from the Minimatch filters supplied.
Downloaded 0.0 MB out of **17,203.8** MB (0%).

liuh-80:
Pipeline issue fixed by this PR: https://github.com/sonic-net/sonic-snmpagent/pull/287

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -85,8 +85,14 @@ def update_data(self):
             if not ent:
                 continue
 
-            # Example output: oid:0x3a000000000608
-            bridge_port_id = ent["SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID"][6:]
+            bridge_port_id = ""
+            try:
+                # Example output: oid:0x3a000000000608
+                bridge_port_id = ent["SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID"][6:]
```
qiluo-msft, line 91: >ent [](http://example.com/codeflow?start=33&length=3)

There are two `[]` on this line, which may throw? only catch that specific operation.

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -85,8 +85,14 @@ def update_data(self):
...
+            except:
```
qiluo-msft, line 92: >except [](http://example.com/codeflow?start=12&length=6)

Catch more specific exception, or use if-else to check.

liuh-80: Fixed, only catch KeyError

src/sonic_ax_impl/mibs/ietf/rfc4363.py
liuh-80: Fixed

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -85,8 +85,15 @@ def update_data(self):
             if not ent:
                 continue
 
+            bridge_port_id_attr = ""
+            try:
+                bridge_port_id_attr = ent["SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID"]
+            except KeyError as e:
+                mibs.logger.error("SyncD 'ASIC_DB' includes invalid FDB_ENTRY '{}': failed to get bridge_port_id, exception: {}".format(fdb_str, e))
+                continue
```
qiluo-msft, line 93: >continue [](http://example.com/codeflow?start=16&length=8)

If we continue and it will the error message will repeat. In a real device, how frequent is the same error messages? do we need to damp the speed?

liuh-80: The log will repeat every 5 seconds, seems not necessary to add more delay:

in https://github.com/sonic-net/sonic-snmpagent/blob/master/src/sonic_ax_impl/main.py

# Background task update frequency ( in seconds )
DEFAULT_UPDATE_FREQUENCY = 5

liuh-80: Fixed, every FDB only output error log once.

qiluo-msft: Thanks for the fix! Further suggestion: flush `broken_fdbs` in `reinit_data()` in order to prevent it increasing size forever in case the FDB entries become crazy.

src/sonic_ax_impl/mibs/ietf/rfc4363.py
```diff
@@ -85,8 +86,18 @@ def update_data(self):
             if not ent:
                 continue
 
+            bridge_port_id_attr = ""
+            try:
+                bridge_port_id_attr = ent["SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID"]
+            except KeyError as e:
+                # Only write error log once
+                if fdb_str not in self.broken_fdbs:
+                    mibs.logger.error("SyncD 'ASIC_DB' includes invalid FDB_ENTRY '{}': failed to get bridge_port_id, exception: {}".format(fdb_str, e))
```
qiluo-msft, line 95: log at warning level, because we skip it and continue with others.

src/sonic_ax_impl/mibs/ietf/rfc4363.py
liuh-80: Fixed

src/sonic_ax_impl/mibs/ietf/rfc4363.py
liuh-80: Fixed

qiluo-msft (approved):

PR #289, opened 2023-08-23
Fix key missing exception when invalied transiver info in STATE_DB

liuh-80:
Fix PhysicalSensorTableMIBUpdater crash when 'type' attribute missing.

#### Work item tracking
Microsoft ADO (number only): 24869277

**- What I did**
Fix PhysicalSensorTableMIBUpdater crash when 'type' attribute missing.

**- How I did it**
Check 'type' attribute, if missing write error log in PhysicalSensorTableMIBUpdater.

**- How to verify it**
Manually test.
Pass all UT

Fix PhysicalSensorTableMIBUpdater crash when 'type' attribute missing.

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc3433.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc3433.py
@@ -385,6 +385,7 @@ class PhysicalSensorTableMIBUpdater(MIBUpdater):
         self.ent_phy_sensor_precision_map = {}
         self.ent_phy_sensor_value_map = {}
         self.ent_phy_sensor_oper_state_map = {}
+        self.broken_transceiver_info.clear()
 
         transceiver_dom_encoded = Namespace.dbs_keys(self.statedb, mibs.STATE_DB, self.TRANSCEIVER_DOM_KEY_PATTERN)
         if transceiver_dom_encoded:
@@ -423,18 +424,17 @@ class PhysicalSensorTableMIBUpdater(MIBUpdater):
                      in STATE_DB, skipping".format(transceiver_dom_entry))
                 continue
 
-            # skip RJ45 port
             transceiver_info_entry_data = Namespace.dbs_get_all(self.statedb, mibs.STATE_DB, mibs.transceiver_info_table(interface))
-    
             if 'type' not in transceiver_info_entry_data:
                 # Only write error log once
                 if interface not in self.broken_transceiver_info:
-                    mibs.logger.error(
+                    mibs.logger.warn(
                         "Invalid interface {} in STATE_DB, \
                         attribute 'type' missing in transceiver_info '{}'".format(interface, transceiver_info_entry_data))
                     self.broken_transceiver_info.append(interface)
                 continue
 
+            # skip RJ45 port
             if  transceiver_info_entry_data['type'] == RJ45_PORT_TYPE:
                 continue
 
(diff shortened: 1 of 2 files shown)


src/sonic_ax_impl/mibs/ietf/rfc3433.py
```diff
@@ -424,6 +425,16 @@ def update_xcvr_dom_data(self):
 
             # skip RJ45 port
```
qiluo-msft, line 426: ># skip RJ45 port [](http://example.com/codeflow?start=12&length=16)

Better to move comments near the line of `RJ45_PORT_TYPE`.

src/sonic_ax_impl/mibs/ietf/rfc3433.py
```diff
@@ -424,6 +425,16 @@ def update_xcvr_dom_data(self):
...
             transceiver_info_entry_data = Namespace.dbs_get_all(self.statedb, mibs.STATE_DB, mibs.transceiver_info_table(interface))
+    
+            if 'type' not in transceiver_info_entry_data:
+                # Only write error log once
+                if interface not in self.broken_transceiver_info:
+                    mibs.logger.error(
```
qiluo-msft, line 432: >error [](http://example.com/codeflow?start=32&length=5)

log at warning level, because it is tolerable to snmpagent.

src/sonic_ax_impl/mibs/ietf/rfc3433.py
```diff
@@ -424,6 +425,16 @@ def update_xcvr_dom_data(self):
...
+                        "Invalid interface {} in STATE_DB, \
+                        attribute 'type' missing in transceiver_info '{}'".format(interface, transceiver_info_entry_data))
+                    self.broken_transceiver_info.append(interface)
```
qiluo-msft, line 435: >broken_transceiver_info [](http://example.com/codeflow?start=25&length=23)

Not an issue and just comment: since the `broken_transceiver_info` is indexed by `interface`, and `interface` is validated above, so there are limited number of possible `interface`, so the `broken_transceiver_info` will have size limitation.

src/sonic_ax_impl/mibs/ietf/rfc3433.py
liuh-80: Fixed

src/sonic_ax_impl/mibs/ietf/rfc3433.py
liuh-80: Fixed

src/sonic_ax_impl/mibs/ietf/rfc3433.py
liuh-80: Fixed by flush broken_transceiver_info in reinit_data()

qiluo-msft: Sorry for misleading, my point is this is not an issue or a bug. Because `broken_transceiver_info` has limited size by its current usage. No need to reinit.

liuh-80: Unnecessary change reverted.

qiluo-msft (approved):

ZhaohuiS:
@liuh-80 Does this change need to be cherry picked into 202012, 202205, 202305?

StormLiangMS:
@liuh-80  @qiluo-msft do we also need this on 202012?

qiluo-msft:
No such bug on 202111 or earlier branches. This PR is needed on 202205, 202305, but these branches are not existing right now.

liuh-80:
> @liuh-80 @qiluo-msft do we also need this on 202012?

We don't need this on 202012 because the code which has issue does not exist on 202012.

PR #303, opened 2023-12-20
Fix SNMP dropping some of the queue counter when create_only_config_d…

DavidZagury:
…b_buffers is set to true

This happened because the MIB assumed that half the queues configured are for mcast. When create_only_config_db_buffers is set to true this is not the case

**- What I did**
The ciscoSwitchQosMIB MIB assumed that all the counter are configured and that half of the configured queues are for mcast.
This is no longer true, the feature "polling only configured ports buffer queue" make it possible for port to not have MC counters.
This wrong assumption caused issue https://github.com/sonic-net/sonic-buildimage/issues/17448

To fix this, I instead used the BUFFER_MAX_PARAM_TABLE to find the max possible queues
**- How I did it**

**- How to verify it**
Inside the SNMP docker run snmp walk:
$ snmpwalk -v2c -c msft 10.64.247.240 1.3.6.1.4.1.9.9.580.1.5.5.1.4.1
Check that the results is not missing counters from any queue


dprital:
@qiluo-msft , Can you please review ?

qiluo-msft:
>             Namespace.get_sync_d_from_all_namespace(mibs.init_sync_d_queue_tables, self.db_conn)

I see a bug inside `init_sync_d_queue_tables`. Instead of spliting a queue_name, we can use "COUNTERS_QUEUE_PORT_MAP", "COUNTERS_QUEUE_INDEX_MAP" to get port index, queue index.

---
Refers to: src/sonic_ax_impl/mibs/vendor/cisco/ciscoSwitchQosMIB.py:92 in 3d13de4. [](commit_id = 3d13de4fec13be532aa8f3edc1ef1acd8f8836c0, deletion_comment = False)

tests/mock_tables/counters_db.json
```diff
@@ -3555,7 +3555,13 @@
...
     "Ethernet8:14": "oid:0x15000000010256",
-    "Ethernet8:15": "oid:0x15000000010257"
+    "Ethernet8:15": "oid:0x15000000010257",
+    "Ethernet16:0":  "oid:0x15000000000260",
+    "Ethernet16:1":  "oid:0x15000000000261",
+    "Ethernet16:2":  "oid:0x15000000000262",
+    "Ethernet16:3":  "oid:0x15000000000263",
+    "Ethernet16:4":  "oid:0x15000000000264",
+    "Ethernet16:6":  "oid:0x15000000000266"
   },
```
qiluo-msft, line 3565: Could you also add test data with below situation?
1. an extra port with incomplete MC
2. an extra port with incomplete UC and incomplete MC.

dprital:
@qiluo-msft , Can you please set the label of : "Request for 202311 branch" ?

src/sonic_ax_impl/mibs/vendor/cisco/ciscoSwitchQosMIB.py
```diff
@@ -67,6 +67,10 @@ def __init__(self):
         self.port_index_namespace = {}
         self.namespace_db_map = Namespace.get_namespace_db_map(self.db_conn)
 
+        self.statedb = mibs.init_db()
```
SuvarnaMeenakshi, line 70: This will not work for multi-asic database as this will only initialize the host database.
Could use the  self.db_conn and connect to state_db using:
Namespace.connect_all_dbs(self.db_conn, mibs.STATE_DB)

src/sonic_ax_impl/mibs/vendor/cisco/ciscoSwitchQosMIB.py
```diff
@@ -132,8 +136,10 @@ def update_stats(self):
             namespace = self.port_index_namespace[if_index]
 
             # The first half of queue id is for ucast, and second half is for mcast
-            # To simulate vendor OID, we wrap queues by half distance
-            pq_count = math.ceil((max(if_queues) + 1) / 2)
+            # To simulate vendor OID, we wrap queues by max priority groups
+            port_max_queues = self.statedb.get_all(self.statedb.STATE_DB,
```
SuvarnaMeenakshi, line 140: Will this param be same for all asic namespaces in multi-asic platform?
If it will be same, then this could be read from the first asic namespace.
self.db_conn[1].get_all()
or could use Namespace.dbs_get_all(self.db_conn..) -- this will return a dict with data from all namespaces.

tests/test_queues_stat.py
```diff
@@ -44,6 +44,51 @@ def test_getQueueCounters(self):
             self.assertEqual(str(value0.name), str(oid))
             self.assertEqual(value0.data, 1)
```
SuvarnaMeenakshi, line 46: can you add a similar test for multi-asic as well:
tests/namespace/test_queues_stat.py

src/sonic_ax_impl/mibs/vendor/cisco/ciscoSwitchQosMIB.py
```diff
@@ -132,8 +136,10 @@ def update_stats(self):
...
+                                                   mibs.buffer_max_parm_table(self.oid_name_map[if_index]))['max_queues']
+            pq_count = math.ceil(int(port_max_queues) / 2)
```
SuvarnaMeenakshi, line 142: Why is this halving done when now? There could be a possibility of MC queue not present, right?

DavidZagury: Yes, there is a possibility of MC queue not present.
We don't want to change the behavior, the purpose of  the halving here is to send the MC counter no on 9-16 rather on 1-8.
If there are MCs - they will be added to the same OID as before, and if there are no MCs, the following for loop will not reach a queue higher then pq_count

DavidZagury:
> > ```
> >         Namespace.get_sync_d_from_all_namespace(mibs.init_sync_d_queue_tables, self.db_conn)
> > ```
> 
> I see a bug inside `init_sync_d_queue_tables`. Instead of spliting a queue_name, we can use "COUNTERS_QUEUE_PORT_MAP", "COUNTERS_QUEUE_INDEX_MAP" to get port index, queue index.
> 
> Refers to: src/sonic_ax_impl/mibs/vendor/cisco/ciscoSwitchQosMIB.py:92 in [3d13de4](https://github.com/sonic-net/sonic-snmpagent/commit/3d13de4fec13be532aa8f3edc1ef1acd8f8836c0). [](commit_id = [3d13de4](https://github.com/sonic-net/sonic-snmpagent/commit/3d13de4fec13be532aa8f3edc1ef1acd8f8836c0), deletion_comment = False)

Hi @qiluo-msft  - I don't see what is the issue you are seeing here. And I don't know how could we get the port index only by using these tables withouth COUNTERS_QUEUE_NAME_MAP or at lest COUNTERS_PORT_NAME_MAP.
Any way, this code is not touched by this PR, if there is an issue with this, shouldn't we address it in a different PR?

dgsudharsan:
@qiluo-msft  @StormLiangMS  Can you please add label for 202305?

dprital:
@qiluo-msft , can you please review and approve ?

qiluo-msft:
>             Namespace.get_sync_d_from_all_namespace(mibs.init_sync_d_queue_tables, self.db_conn)

@DavidZagury this comment is still applicable. Will you fix it in this PR?

---
In reply to: [1869825240](https://github.com/sonic-net/sonic-snmpagent/pull/303#issuecomment-1869825240) [](http://example.com/codeflow?ancestors=1869825240)

---
Refers to: src/sonic_ax_impl/mibs/vendor/cisco/ciscoSwitchQosMIB.py:92 in 3d13de4. [](commit_id = 3d13de4fec13be532aa8f3edc1ef1acd8f8836c0, deletion_comment = False)

qiluo-msft (approved):
LGTM, please check with other active reviewers.

tests/mock_tables/global_db/state_db.json
```diff
@@ -123,5 +123,32 @@
   "TEMPERATURE_INFO|thermal1": {
     "temperature": "20.5",
     "is_replaceable": "False"
+  },
+  "BUFFER_MAX_PARAM_TABLE|Ethernet0": {
```
SuvarnaMeenakshi, line 127: state_db in global db will not have interface related configuration

src/sonic_ax_impl/mibs/vendor/cisco/ciscoSwitchQosMIB.py
SuvarnaMeenakshi: what will be the value of port_max_queues if there are Unicast and multicast queues ?

StormLiangMS:
@dgsudharsan @DavidZagury  could you address comment from Suvarna?

StormLiangMS:
@DavidZagury  @dgsudharsan pls test this PR with 202305 image.

StormLiangMS:
ADO:26158684

tests/mock_tables/global_db/state_db.json
DavidZagury: Moved to asics db

DavidZagury:
> @DavidZagury @dgsudharsan pls test this PR with 202305 image.

I tested it on 202305

SuvarnaMeenakshi (approved):

dgsudharsan:
@qiluo-msft @SuvarnaMeenakshi Can you please merge this?

qiluo-msft (approved):

DavidZagury:
@yxieca can we merge this to 202311?

PR #306, opened 2024-01-10
Set the execute bit on sysDescr_pass.py

AKosturArista:
#### Why I did it
Slim images may have a sysDescr_pass.py script that does not correctly have the execute bit set resulting in any request for the sysDescr.0 OID via SNMP would fail.  By setting the execute bit directly in source control, this script is no longer reliant on any deployment script to fix the permissions after the fact.

#### How I did it
Set the execute bit directly within source control

#### How to verify it
Testing on a platform which uses a slim image should reliably pass the snmp/test_snmp_v2mib.py test.

#### Which release branch to backport (provide reason below if selected)
 * [ ]  201811
 * [ ]  201911
 * [ ]  202006
 * [ ]  202012
 * [ ]  202106
 * [ ]  202111
 * [ ]  202205
 * [ ]  202211
 * [x]  202305


SuvarnaMeenakshi:
This seems like a work around, because with the slim image loaded, I was able to run a python script inside snmp docker to copy the sysDescr_pass.py to the location and change the mode to allow execution. What is the reason that we are able to do it during run time but does not occur when sonic_ax_impl comes up?

StormLiangMS (approved):
LGTM

StormLiangMS:
hi @SuvarnaMeenakshi I plan to take this as a workaround for now.

1. only one file get impacted by this permission missing issue, other files which are impacted are not run time using. more details in https://github.com/aristanetworks/sonic-qual.msft/issues/46
2. this issue could be caused by the build of process of slim image, it may take longer time to figure out, arista can keep going, but it won't be a blocker issue for slim images. 

let me know if any concern.

SuvarnaMeenakshi (approved):
Approving the work around for slim image fix.

PR #312, opened 2024-03-11
Updating SNMP implementation to handle change of PSU Keys

gregoryboudreau:
**- What I did**
Modified snmp implementation of Cisco MIB to use updated format for psu keys instead of index of PSU
**- How I did it**
Used a consistently sorted list of keys to allow for the modified statedb key naming for psus
**- How to verify it**
Ran the associated sonic-mgmt tests for PSUs (phy_entity and snmp_psu)

Modifies SNMP implementation of Cisco PSUs to use key names instead of simple indexing

Related PRs:
https://github.com/sonic-net/sonic-mgmt/pull/11944
https://github.com/sonic-net/sonic-utilities/pull/3208
https://github.com/sonic-net/sonic-platform-daemons/pull/446


gechiang (approved):

abdosi:
let's fix this and rerun,

src/sonic_ax_impl/mibs/vendor/cisco/ciscoEntityFruControlMIB.py
```diff
@@ -70,7 +71,8 @@ def _get_psu_presence(self, psu_index):
         Get PSU presence
         :return: the presence of particular PSU
         """
-        psu_name = PSU_INFO_KEY_TEMPLATE.format(psu_index)
+        psu_keys = natsorted(self.statedb.keys(self.statedb.STATE_DB, "PSU_INFO|*"))
+        psu_name = psu_keys[psu_index-1].removeprefix("PSU_INFO|")
```
SuvarnaMeenakshi, line 75: minor comment, in line, 75 psu_name is actually not "psu_name",  but it is the numerical value from psu name and is then used to get the psu key mibs.psu_info_table(psu_name).
Can we rename "psu_name" in line 75 or directly use "psu_keys[psu_index-1]" in line 75 instead of mibs.psu_info_table(psu_name)

gregoryboudreau: you are correct in that the `mibs.psu_info_table` is redundant as we already have the full key for redis, I will update this accordingly.

SuvarnaMeenakshi:
@gregoryboudreau , will there by any impact in rfc2737 implementation which uses PSU_INFO?
https://github.com/sonic-net/sonic-snmpagent/blob/25f9e4fa16e8f4789f770b4c93537bb2026c8ec5/src/sonic_ax_impl/mibs/ietf/rfc2737.py#L869

SuvarnaMeenakshi:
@gregoryboudreau can you merge with latest master

gregoryboudreau:
> @gregoryboudreau , will there by any impact in rfc2737 or 3433 implementation which uses PSU_INFO?
> 
> https://github.com/sonic-net/sonic-snmpagent/blob/25f9e4fa16e8f4789f770b4c93537bb2026c8ec5/src/sonic_ax_impl/mibs/ietf/rfc2737.py#L869

definitely possible, didn't see that originally. Will take a look.

mlok-nokia (approved):

gregoryboudreau:
> > @gregoryboudreau , will there by any impact in rfc2737 or 3433 implementation which uses PSU_INFO?
> > https://github.com/sonic-net/sonic-snmpagent/blob/25f9e4fa16e8f4789f770b4c93537bb2026c8ec5/src/sonic_ax_impl/mibs/ietf/rfc2737.py#L869
> 
> definitely possible, didn't see that originally. Will take a look.

Took a look and I don't think it will cause any problem. The logic seems to just plug into `PSU_INFO|*` which will still get the keys as expected. The key pattern is still the same in that all PSUs will be under the same table, just that their key (what would be the *) is no longer just an enumerated PSU name but the actual vendor defined name (if available)

SuvarnaMeenakshi (approved):
lgtm

SuvarnaMeenakshi:
hi @gregoryboudreau do we have to follow an order for merge of PRs in the various repos or are the independent of each other?

gregoryboudreau:
> hi @gregoryboudreau do we have to follow an order for merge of PRs in the various repos or are the independent of each other?

everything but platform-daemons can go in in any order, platform-daemons must be last

PR #318, opened 2024-05-06
fixes snmp not to use blocking get_all() after keys() 

pttch:
**- What I did**
In a dynamic environment, it is possible that some of the keys may disappear between invoking keys() and get_all().
Prevent unnecessary timeout of blocking get_all(). Add more snmp no blocking fix for https://github.com/sonic-net/sonic-snmpagent/pull/255

**- How I did it**
  Modify the blocking mode from True to False

**- How to verify it**
   Use server to send with 1000 requests/second. Keep test for weekend and it do not happen.

Prevent unnecessary timeout of blocking get_all(). Add more snmp no blocking fix.


pttch:
Modify to use correct email for commit

PR #322, opened 2024-07-03
[ciscoPfcExtMIB]: Remove returning first intf index if subid is empty

SuvarnaMeenakshi:
**- What I did**
When querying for PFC counters MIB cpfcIfTable if there is no interface present on the device, then there is an exception while handing get_next as the implementation tries to return the first index of an empty list.
This happens in case of VoQ chassis Supervisor where there are no interfaces present. 
Anytime a snmp query is made to cpfcIfTable 1.3.6.1.4.1.9.9.813.1.1, the exception message gets logged on the device.

**- How I did it**
Currently if sub_id is None, if_range[0] is returned. IF if_range is empty array this can cause an Exception.
Removed this check to avoid returning if_range[0].
The check "if not sub_id" is not required here.
right = bisect_right(self.if_range, sub_id) will handle if the sub_id is empty or if the if_range is empty by returning 0 index.

**- How to verify it**
Verified that no exception is logged upon query in VoQ chassis supervisor.
Other platforms like Linecard, single-asic device - no Change in the SNMP result.

### DIFF
--- a/src/sonic_ax_impl/mibs/vendor/cisco/ciscoPfcExtMIB.py
+++ b/src/sonic_ax_impl/mibs/vendor/cisco/ciscoPfcExtMIB.py
@@ -70,9 +70,6 @@ class PfcUpdater(MIBUpdater):
         :return: the next sub id.
         """
         try:
-            if not sub_id:
-                return self.if_range[0]
-
             right = bisect_right(self.if_range, sub_id)
             if right >= len(self.if_range):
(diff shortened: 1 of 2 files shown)


abdosi (commented):
is it possible to add/update  UT  for this ?

abdosi (approved):
Lgtm

qiluo-msft (approved):

PR #324, opened 2024-07-12
Modify path of python-wheels package to use bookworm

SuvarnaMeenakshi:
**- What I did**
Modify path of python-wheels package to use bookworm.
Seeing the below errors in azure pipeline:
```
WARNING: Requirement '../target/python-wheels/buster/swsssdk*-py3-*.whl' looks like a filename, but the file does not exist
ERROR: swsssdk*-py3-*.whl is not a valid wheel filename.
```
**- How I did it**
Modify path of python-wheels package to use bookworm
**- How to verify it**

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -69,7 +69,7 @@ stages:
         set -ex
         # Install .NET CORE
         curl -sSL https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -
-        sudo apt-add-repository https://packages.microsoft.com/debian/10/prod
+        sudo apt-add-repository https://packages.microsoft.com/debian/12/prod
         sudo apt-get update
         sudo apt-get install -y dotnet-sdk-8.0
       displayName: "Install .NET CORE"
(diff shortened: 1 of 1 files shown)


saiarcot895:
Can you also bump up the debian version used in the pipeline to bookworm?

azure-pipelines.yml
```diff
@@ -52,8 +52,8 @@ stages:
         sudo apt-get -y purge libhiredis-dev libnl-3-dev libnl-route-3-dev
         sudo apt-get -y install libhiredis0.14
         sudo dpkg -i ../target/debs/buster/{libyang_1.0.73_amd64.deb,libswsscommon_1.0.0_amd64.deb,python3-swsscommon_1.0.0_amd64.deb,libnl-3-200_*.deb,libnl-genl-3-200_*.deb,libnl-nf-3-200_*.deb,libnl-route-3-200_*.deb}
```
saiarcot895, line 54: This needs to be changed to use bookworm instead of buster.

saiarcot895 (approved):

PR #343, opened 2024-12-16
Fix redis memory leak issue in PhysicalEntityCacheUpdater

yejianquan:
**- What I did**
Fixes the redis memory leak bug:
https://github.com/sonic-net/sonic-snmpagent/issues/342
There's chance that the physical_entity_updater creates subscriptions to redis, and never consume the messages due to exceptions.
Then the memory buffer(omem) of redis client starts to increase and never end, redis memory leaks.

The reason is all 5 physical entity cache updaters inherit from PhysicalEntityCacheUpdater.
In the first update_data, they initialize the psubscription to redis database.
https://github.com/sonic-net/sonic-snmpagent/blob/c5301b2cf94304deb8e09959d67af6a089155507/src/sonic_ax_impl/mibs/ietf/rfc2737.py#L667
And everytime when the update_data is called again, it get the message from the psub and process.
https://github.com/sonic-net/sonic-snmpagent/blob/c5301b2cf94304deb8e09959d67af6a089155507/src/sonic_ax_impl/mibs/ietf/rfc2737.py#L678

And outside, in the logic of the MIBUpdater, it calls update_data more frequently than reinit_data.
A side-effect is, if reinit_data failed forever, the update_counter will not been cleaned, then update_data will not be called forever.
https://github.com/sonic-net/sonic-snmpagent/blob/c5301b2cf94304deb8e09959d67af6a089155507/src/ax_interface/mib.py#L43

So the problem is, at the begining, the psub is created at the first update_data and all things work well, until an unrecoverable issue happened, 
`PHYSICAL_ENTITY_INFO|PSU *` missing in the database (it's a pmon issue)
This causes both reinit_data and update_data to be failed, because all of them finally call _update_per_namespace_data, which tries to cast an empty string '' to int and raises ValueError.
Then the update_data is not called forever, because reinit_data will never success.
But the previously established psubscription is still there, and no one gonna to consume it(the update_data is blocked), then Redis database memory starts to slowly leak.

**- How I did it**
1. Catch the exception during the loop of reinit_data, make sure the reinit_data of every physical_entity_updater will be called
2. Clear message and cancel the subscription in the reinit_data, avoid the message accumulates in the redis subscription queue

**- How to verify it**
Tested on Cisco chassis, the memory is not leaking anymore.

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc2737.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc2737.py
@@ -318,8 +318,34 @@ class PhysicalTableMIBUpdater(MIBUpdater):
         self.physical_name_map[chassis_mgmt_sub_id] = name
         self.physical_fru_map[chassis_mgmt_sub_id] = self.NOT_REPLACEABLE
 
+        exceptions = []
+        has_runtime_err = False
+        # Catch exception in the iteration
+        # This makes sure if any exception is raised in the mid of loop
+        # every updater's reinit_data function will be always called
+        # So that the redis subscriptions always get chance to be cleaned,
+        # Otherwise if the exception never recover,
+        # the redis subscription keeps increasing but never got consumed,
+        # this causes redis memory leak.
         for updater in self.physical_entity_updaters:
-            updater.reinit_data()
+            try:
+                updater.reinit_data()
+            except BaseException as e:
+                mibs.logger.error(f"Jianquan {e}, type{type(e)}")
+                if isinstance(e, RuntimeError):
+                    has_runtime_err = True
+                # Log traceback so that we know the original error details
+                mibs.logger.error(e, exc_info=True)
+                exceptions.append(e)
+
+        # The RuntimeError will be considered as Redis connection error
+        # And will trigger re-init connection, if the exceptions contain any RuntimeError
+        # We raise runtime error
+        if exceptions:
+            if has_runtime_err:
+                raise RuntimeError(exceptions)
+            else:
+                raise Exception(exceptions)
 
     def update_data(self):
         # This code is not executed in unit test, since mockredis
@@ -648,6 +674,21 @@ class PhysicalEntityCacheUpdater(object):
         self.entity_to_oid_map = {}
 
     def reinit_data(self):
+
+        # Redis subscriptions are established and consumed in update_data,
+        # but if there's stable exception during update logic,
+        # the reinit_data will be called, but the update_data is never called.
+        # The message is sent into subscription queue, but never got consumed,
+        # this causes Redis memory leaking.
+        # Hence clear the message in the subscription and cancel the subscription during reinit_data
+        for db_index in list(self.pub_sub_dict):
+            pubsub = self.pub_sub_dict[db_index]
+            db_conn = self.mib_updater.statedb[db_index]
+            # clear message in the subscription and cancel the subscription
+            mibs.clear_pubsub_msg(pubsub)
+            mibs.cancel_redis_pubsub(pubsub, db_conn, db_conn.STATE_DB, self.get_key_pattern())
+            del self.pub_sub_dict[db_index]
+
         self.entity_to_oid_map.clear()
         # retrieve the initial list of entity in db
         key_info = Namespace.dbs_keys(self.mib_updater.statedb, mibs.STATE_DB, self.get_key_pattern())
--- a/src/sonic_ax_impl/mibs/ietf/rfc2737.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc2737.py
@@ -331,7 +331,6 @@ class PhysicalTableMIBUpdater(MIBUpdater):
             try:
                 updater.reinit_data()
             except BaseException as e:
-                mibs.logger.error(f"Jianquan {e}, type{type(e)}")
                 if isinstance(e, RuntimeError):
                     has_runtime_err = True
                 # Log traceback so that we know the original error details
--- a/.gitignore
+++ b/.gitignore
@@ -136,3 +136,4 @@ fabric.properties
 
 gh-release.patch
 tests/test_cpuUtilizationHandler.py
+tests/test-results.xml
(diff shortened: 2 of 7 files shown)


yejianquan:
@abdosi @sdszhang @sdszhang @vperumal @anamehra for viz

yejianquan:
Removed all the sonicbld comment due to adding new test cases

yejianquan:
Hi @Junchao-Mellanox  @liuh-80  @SuvarnaMeenakshi , could you help to review this PR?

yejianquan:
@qiluo-msft for viz

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -648,6 +674,21 @@ def __init__(self, mib_updater):
...
+        # but if there's stable exception during update logic,
+        # the reinit_data will be called, but the update_data is never called.
+        # The message is sent into subscription queue, but never got consumed,
+        # this causes Redis memory leaking.
+        # Hence clear the message in the subscription and cancel the subscription during reinit_data
+        for db_index in list(self.pub_sub_dict):
+            pubsub = self.pub_sub_dict[db_index]
+            db_conn = self.mib_updater.statedb[db_index]
+            # clear message in the subscription and cancel the subscription
+            mibs.clear_pubsub_msg(pubsub)
```
SuvarnaMeenakshi, line 688: Seems like the pub_sub_dict is getting cleared upon every reinit_data() call, is that the intention or should the clearing be done only upon exception?

src/sonic_ax_impl/mibs/ietf/rfc2737.py
```diff
@@ -318,8 +318,34 @@ def reinit_data(self):
...
+        # So that the redis subscriptions always get chance to be cleaned,
+        # Otherwise if the exception never recover,
+        # the redis subscription keeps increasing but never got consumed,
+        # this causes redis memory leak.
         for updater in self.physical_entity_updaters:
-            updater.reinit_data()
+            try:
+                updater.reinit_data()
+            except BaseException as e:
+                mibs.logger.error(f"Jianquan {e}, type{type(e)}")
```
SuvarnaMeenakshi, line 334: type on error log

yejianquan: Debug log for unit test, removed, thanks for pointing out!

src/sonic_ax_impl/mibs/ietf/rfc2737.py
yejianquan: Hi @SuvarnaMeenakshi , yes, I intentionally design to clear the subscription on every reinit_data() call. Combined with the external exception catch, we can confirm the subscription will always be cleared.

As for the logic that only clear subscriptions on the exceptions, definitely it makes sense, but there's trade off
if we **keep current logic**
### Adv:
####  We can confirm no memory leaking is expected, because every reinit_data clears the subscriptions.
### Disadv:
####  We re-establish the psubscription to redis server every 60s.

After some research, the psubscription establishment is not heavy and generally considered as light weight.
The heavy works are the message sending after psub established.

If we **only clear on exception**
### Adv:
#### We don't need to re-establish the psubscriptions.
### Disadv:
#### Only when exceptions happened we clear the psubscriptions. If there's no exception raised, the psubscriptions will not be cleared.

From my understanding, semantically, the cleaning of the psub is not a directly result of catching an exception, it's like a patched bug fix saying we know if exception got raised, there is **possibility** that the data in the psub doesn't get consumed(But it could be consumed actually before the exception), so I clean it no matter what the exception is.

But in current approach, I prefer to consider reinit_data a scheduled total refresh, like what it does to the entity caches, clean all of them and fetch from the database no matter got notified by the psub or not. So in the total refresh, we clean the subs, don't sounds like a patched bug fix but a designed feature.
Definitely it would be better if there's a hook specifically to re-initialized all the thing. 
But the reinit_connection only happens on RuntimeError, and I don't want to add new hooks which is a big change to the structure.

@SuvarnaMeenakshi kindly let me know your suggestions and considerations, I will ok with both approaches.

SuvarnaMeenakshi: > Hi @SuvarnaMeenakshi , yes, I intentionally design to clear the subscription on every reinit_data() call. Combined with the external exception catch, we can confirm the subscription will always be cleared.
> 
> As for the logic that only clear subscriptions on the exceptions, definitely it makes sense, but there's trade off if we **keep current logic**
> 
> ### Adv:
> #### We can confirm no memory leaking is expected, because every reinit_data clears the subscriptions.
> ### Disadv:
> #### We re-establish the psubscription to redis server every 60s.
> After some research, the psubscription establishment is not heavy and generally considered as light weight. The heavy works are the message sending after psub established.
> 
> If we **only clear on exception**
> 
> ### Adv:
> #### We don't need to re-establish the psubscriptions.
> ### Disadv:
> #### Only when exceptions happened we clear the psubscriptions. If there's no exception raised, the psubscriptions will not be cleared.
> From my understanding, semantically, the cleaning of the psub is not a directly result of catching an exception, it's like a patched bug fix saying we know if exception got raised, there is **possibility** that the data in the psub doesn't get consumed(But it could be consumed actually before the exception), so I clean it no matter what the exception is.
> 
> But in current approach, I prefer to consider reinit_data a scheduled total refresh, like what it does to the entity caches, clean all of them and fetch from the database no matter got notified by the psub or not. So in the total refresh, we clean the subs, don't sounds like a patched bug fix but a designed feature. Definitely it would be better if there's a hook specifically to re-initialized all the thing. But the reinit_connection only happens on RuntimeError, and I don't want to add new hooks which is a big change to the structure.
> 
> @SuvarnaMeenakshi kindly let me know your suggestions and considerations, I will ok with both approaches.

Thank you for the details, I understand the approach and sounds like a good way since it is not a heavy operation.

SuvarnaMeenakshi (approved):

PR #345, opened 2025-01-02
Fix snmp agent not-responding issue when high CPU utilization

yejianquan:
Fix https://github.com/sonic-net/sonic-buildimage/issues/21314
The SNMP agent and MIB updaters are basic on Asyncio/Coroutine,
  the mib updaters share the same Asyncio event loop with the SNMP agent client.
  Hence during the updaters executing, the agent client can't receive/respond to new requests.

When the CPU utilization is high (In some stress test we make CPU 100% utilization), the updates are slow, and this causes the snmpd request to be timeout because the agent got suspended during updating.

**- What I did**
Decrease the MIB update frequency when the update execution is slow.
pros:
   1.  The snmp request can success even if 100% CPU utilization.
   2. The snmpd request seldomly fails due to timeout, combined with https://github.com/sonic-net/sonic-buildimage/pull/21316 , we have 4*5 = 20s time windows for the SNMP agent to wait for the MIB updates finish and respond to snmpd request. 
   3. Relief the CPU cost when CPU is high, the can avoid CPU becomes more crowded.

cons:
   1. Tested on pizzabox (4600c), the updaters are very fast, generally finished within 0.001~0.02s, the chagne won't actually affect the frequency and interval.
   2. On Cisco chassis, the update of SNMP data could be delayed for 10~20s(at most 60s in extreme situation). Per my oberservation, most of the updater finishes within 0.5s. But for
         1.a ciscoSwitchQosMIB.QueueStatUpdater, generally finishs in 0.5-2s, expected to be delayed to 5-20s
         1.b PhysicalTableMIBUpdater, generally finishs in 0.5-1.5s, expected to be delayed to 5-1.5s
         1.c ciscoPfcExtMIB.PfcPrioUpdater, generally finishs in 0.5-3s, expected to be delayed to 5-30s
       

**- How I did it**
In get_next_update_interval, we compute the interval based on current execution time.
Roughly, we make the 'update interval'/'update execution time' >= UPDATE_FREQUENCY_RATE(10)
More specifically, if the execution time is 2.000001s, we sleep 21s before next update round.
And the max interval won't be longer than MAX_UPDATE_INTERVAL(60s).

**- How to verify it**
Test on Cisco chassis, test_snmp_cpu.py which triggers 100% CPU utilization test whether snmp requests work well.

### DIFF
--- a/src/ax_interface/agent.py
+++ b/src/ax_interface/agent.py
@@ -8,7 +8,7 @@ BACKGROUND_WAIT_TIMEOUT = 10  # seconds
 
 
 class Agent:
-    def __init__(self, mib_cls, update_frequency, loop):
+    def __init__(self, mib_cls, enable_dynamic_frequency, update_frequency, loop):
         if not type(mib_cls) is MIBMeta:
             raise ValueError("Expected a class with type: {}".format(MIBMeta))
 
@@ -20,7 +20,7 @@ class Agent:
         self.stopped = asyncio.Event()
 
         # Initialize our MIB
-        self.mib_table = MIBTable(mib_cls, update_frequency)
+        self.mib_table = MIBTable(mib_cls, enable_dynamic_frequency, update_frequency)
 
         # containers
         self.socket_mgr = SocketManager(self.mib_table, self.run_enabled, self.loop)
--- a/src/ax_interface/constants.py
+++ b/src/ax_interface/constants.py
@@ -90,3 +90,8 @@ class PduTypes(int, Enum):
 
 
 DEFAULT_PDU_TIMEOUT = 5
+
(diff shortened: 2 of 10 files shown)


yejianquan:
Hi @SuvarnaMeenakshi , could you help to review this PR?
The only concern is the potential delay of SNMP data, described in cons.
Do you think the delay would result in risk in production env?

yejianquan:
FYI @abdosi @vperumal  @anamehra

yejianquan:
Hi @qiluo-msft  @SuvarnaMeenakshi 
Limited the scope of dynamic frequency, please help to review again, thanks!

qiluo-msft (approved):

PR #368, opened 2026-03-10
rfc2737: check if entry is not empty before accessing it

ymd-arista:
In `PsuCacheUpdater` and `FanCacheUpdater`, when we try to get
information of a PSU/Fan from Physical entity info table, we fail to check
whether there is no entry for that psu/fan in the table.

We see this issue in https://github.com/sonic-net/sonic-buildimage/issues/25652, where going
further without checking if there is no entry causes an exception in such
scenarios. This patch aims to fix that issue.

### DIFF
--- a/src/sonic_ax_impl/mibs/ietf/rfc2737.py
+++ b/src/sonic_ax_impl/mibs/ietf/rfc2737.py
@@ -888,6 +888,8 @@ class PsuCacheUpdater(PhysicalEntityCacheUpdater):
             return
 
         psu_relation_info = self.get_physical_relation_info(psu_name)
+        if not psu_relation_info:
+            return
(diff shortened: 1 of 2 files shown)


judyjoseph (dismissed):

ymd-arista:
@judyjoseph : Can you re-review this PR, it was blocked earlier from merging because of coverage.

rlhui:
> @judyjoseph : Can you re-review this PR, it was blocked earlier from merging because of coverage.

@judyjoseph ping again

judyjoseph (approved):

ymd-arista:
Closing and reopening the PR to re-trigger the checks which have been stuck for a week.

PR #371, opened 2026-04-11
Fix test collection failure: add root conftest.py and clean up namespace test imports

yxieca:
### Description
The test scripts under `tests/namespace/` have an incorrect `sys.path.insert` that uses only two levels of `os.path.dirname`, landing in `tests/` instead of the repo root where `src/` lives. This caused `sonic_ax_impl` import failures during pytest collection when file traversal order didn't happen to add `src/` first (as seen with Python 3.13/trixie's pytest).

### Changes
1. **conftest.py** (new) — Adds `src/` to `sys.path` at the root level, ensuring all test files can find `sonic_ax_impl` regardless of collection order.
2. **tests/namespace/test_*.py** — Removes the now-redundant `modules_path`/`sys.path.insert` blocks and unused `os`/`sys` imports from 10 test files.

### Note on CI builder
Switching to trixie slave was attempted but reverted — the pipeline also needs updates for `libhiredis` (0.14 doesn't exist on trixie) and deb/wheel artifact paths (hardcoded to `bookworm/`). Trixie CI migration will be addressed in a follow-up PR.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,18 @@
+"""
+Root conftest.py — ensures the local src/ directory is on sys.path before
+pytest collects any test modules.
+
+On trixie+ builds the wheel build flow is:
+    pip install ".[testing]" → pip uninstall asyncsnmp → python -m pytest
+After uninstall, sonic_ax_impl / ax_interface are no longer in site-packages.
+Individual test files add src/ to sys.path, but tests/mock_tables/dbconnector.py
+(imported during collection) does `import sonic_ax_impl.mibs` before any test
+file executes.  Placing the path fixup here resolves the import for all
+collectors.
+"""
+import os
+import sys
+
+_src = os.path.join(os.path.dirname(__file__), "src")
+if _src not in sys.path:
+    sys.path.insert(0, _src)
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,7 +26,7 @@ stages:
       vmImage: ubuntu-latest
 
     container:
-      image: sonicdev-microsoft.azurecr.io:443/sonic-slave-bookworm:latest
+      image: sonicdev-microsoft.azurecr.io:443/sonic-slave-trixie:latest
 
     steps:
     - checkout: self
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,7 +26,7 @@ stages:
       vmImage: ubuntu-latest
 
     container:
-      image: sonicdev-microsoft.azurecr.io:443/sonic-slave-trixie:latest
+      image: sonicdev-microsoft.azurecr.io:443/sonic-slave-bookworm:latest
 
     steps:
     - checkout: self
--- a/tests/namespace/test_arp.py
+++ b/tests/namespace/test_arp.py
@@ -1,9 +1,5 @@
-import os
-import sys
 import importlib
 
-modules_path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-sys.path.insert(0, os.path.join(modules_path, 'src'))
 
 from unittest import TestCase
 from unittest.mock import patch, mock_open
--- a/tests/namespace/test_bgp.py
+++ b/tests/namespace/test_bgp.py
@@ -1,11 +1,7 @@
-import os
-import sys
 import importlib
 # noinspection PyUnresolvedReferences
 import tests.mock_tables.dbconnector
 
-modules_path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-sys.path.insert(0, os.path.join(modules_path, 'src'))
 
 from unittest import TestCase
 from unittest.mock import patch, mock_open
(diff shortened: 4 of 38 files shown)


conftest.py
```diff
@@ -0,0 +1,18 @@
+"""
+Root conftest.py — ensures the local src/ directory is on sys.path before
+pytest collects any test modules.
+
+On trixie+ builds the wheel build flow is:
+    pip install ".[testing]" → pip uninstall asyncsnmp → python -m pytest
+After uninstall, sonic_ax_impl / ax_interface are no longer in site-packages.
```
saiarcot895, line 7: This behavior is not unique to Trixie, and also applies to Bookworm. The reason for the failure is incorrect.

The actual reason tests are failing is because the tests under `tests/namespace/` have this at the top of the file:

```
modules_path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
sys.path.insert(0, os.path.join(modules_path, 'src')
```

This is meant to add the `src` directory into the import path. However, `os.path.dirname(os.path.dirname(os.path.abspath(__file__)))` leaves you with `/sonic/src/sonic-snmpagent/tests` (inside the slave container), not `/sonic/src/sonic-snmpagent` where the `src` directory actually lives. This means that the logic here is incorrect for all 5 test scripts under this directory. This means that one fix is to add another `os.path.dirname` call to get the parent directory, for those 5 test scripts.

The reason this wasn't seen under Bookworm's version of pytest is likely because of file traversal order. If the tests in the `tests/` directory itself were read/processed/collected first before the `tests/namespace/` directory, then the correct `src` directory would already be added to `sys.path`, and the modules would be found.

Using a top-level `conftest.py` is a better way than handling it in each and every test script. What _might_ be even better might be to get rid of the `src` directory entirely and move the contents up one level, but that might need some packaging changes as well.

However, the individual `sys.path.insert`s in each test script should be removed as well. Otherwise, that's just keeping technical debt/dead code.

saiarcot895 (changes requested):
See above.

Please also move the CI builder to Trixie.

conftest.py
yxieca: Thanks for the detailed analysis! You're right — the root cause is the incorrect `os.path.dirname` nesting in the namespace tests, not trixie-specific. The conftest.py approach is the right fix.

I've pushed a second commit that removes all the dead `sys.path.insert` blocks and unused `os`/`sys` imports from the 10 namespace test files.

Regarding moving the CI builder to Trixie — I'll look into that as a follow-up.

yxieca: Also pushed a third commit switching the CI builder from bookworm to trixie as requested.

qiluo-msft:
could you add the error message you see originally into PR description?

conftest.py
yxieca: Tried switching to trixie but reverted — the pipeline also needs updates for libhiredis (0.14 doesn't exist on trixie) and the deb/wheel artifact paths (hardcoded to bookworm/). Will address the full trixie CI migration in a follow-up PR.

yxieca:
Original failure:

_________________ ERROR collecting tests/namespace/test_bgp.py _________________
ImportError while importing test module '/sonic/src/sonic-snmpagent/tests/namespace/test_bgp.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3/dist-packages/_pytest/python.py:493: in importtestmodule
    mod = import_path(
/usr/lib/python3/dist-packages/_pytest/pathlib.py:587: in import_path
    importlib.import_module(module_name)
/usr/lib/python3.13/importlib/__init__.py:88: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1387: in _gcd_import
    ???
<frozen importlib._bootstrap>:1360: in _find_and_load
    ???
<frozen importlib._bootstrap>:1331: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:935: in _load_unlocked
    ???
/usr/lib/python3/dist-packages/_pytest/assertion/rewrite.py:185: in exec_module
    exec(co, module.__dict__)
tests/namespace/test_bgp.py:5: in <module>
    import tests.mock_tables.dbconnector
tests/mock_tables/dbconnector.py:153: in <module>
    import sonic_ax_impl.mibs
E   ModuleNotFoundError: No module named 'sonic_ax_impl'

saiarcot895:
Can you remove the `sys.path.insert` lines for the following tests as well?

```
tests/test_MIBMeta.py
tests/test_PDUEncodings.py
tests/test_PDUHeader.py
tests/test_agent.py
tests/test_arp.py
tests/test_bgp.py
tests/test_fdb.py
tests/test_forward.py
tests/test_hc_interfaces.py
tests/test_interfaces.py
tests/test_lldp.py
tests/test_mibs.py
tests/test_nexthop.py
tests/test_pfc.py
tests/test_psu.py
tests/test_queues_stat.py
tests/test_rfc1213.py
tests/test_rfc2737.py
tests/test_rfc2863.py
tests/test_rfc3433.py
tests/test_rfc4292.py
tests/test_rfc4363.py
tests/test_sensor.py
tests/test_sn.py
```

saiarcot895 (approved):

